PR #871, opened 2025-06-02
804 feat resolver for historical account balances multicall

### LINKED ISSUE #804: feat: Resolver for Historical Account Balances (multicall)
### User story

As a frontend developer, I want to query historical token balances for specific accounts at a given block number so that I can accurately calculate balance variations over time without having to perform these calculations on the frontend.

### Describe the feature request

Currently, the system only stores the current balance of accounts in the `accountBalance` table. And the token holders data table will need to present variation of the holders balance, the current implementation allows to calculate the variation using the "Transfers" table, but it will cost too much processing and not be performatic, so the recommended solution is to implement a viem multicall to get the balance for each address given the blocknumber and token address.

### Describe solution

Implement a GraphQL resolver that uses viem's multicall to fetch historical token balances for multiple accounts at a specific block number. The resolver will:

1. Accept the following parameters:
   - `addresses`: Array of account addresses to query
   - `blockNumber`: The specific block number to query balances at
   - `daoId`: The DAO identifier to filter relevant tokens

2. Implementation details:
   - Use viem's multicall to efficiently batch balance requests
   - Validate block number is within reasonable range (e.g., not too old)
   - Handle RPC errors gracefully
   - Implement rate limiting to prevent abuse
   - Consider implementing caching for frequently requested historical balances

3. GraphQL schema addition:
```graphql
type HistoricalBalance {
  address: String!
  balance: BigInt!
  blockNumber: Int!
  tokenId: String!
}

type Query {
  getHistoricalBalances(
    addresses: [String!]!
    blockNumber: Int!
    daoId: String!
  ): [HistoricalBalance!]!
}
```

4. Performance considerations:
   - Implement pagination if needed for large address lists
   - Add appropriate timeout handling
   - Consider implementing a caching layer for frequently requested data
   - Add rate limiting to prevent abuse

5. Error handling:
   - Validate block number is not too old
   - Handle RPC errors gracefully
   - Provide clear error messages for invalid inputs
   - Implement proper logging for debugging

The solution will:
- Provide consistent and accurate historical balance data
- Improve frontend performance
- Reduce frontend complexity
- Enable more accurate historical analysis

Note: Consider implementing a caching mechanism for frequently requested historical balances to improve performance and reduce RPC calls. Also, consider adding a maximum limit to the number of addresses that can be queried in a single request to prevent abuse and ensure optimal performance.

lgahdl:
Closes #804 
feat: Creating the REST endpoint for historical balances;
feat: Adding deployment for Multicall3 contract in the local node;
refactor: Moving ENS contracts to an "ens" folder in the local node folder structure;
feat: historical balances graphql resolver;

### DIFF
--- a/apps/indexer/ponder.schema.ts
+++ b/apps/indexer/ponder.schema.ts
@@ -205,7 +205,11 @@ export const accountBalanceRelations = relations(accountBalance, ({ one }) => ({
   delegatedTo: one(accountPower, {
     fields: [accountBalance.delegate],
     references: [accountPower.accountId],
-    relationName: "delegatedFromTo",
+  }),
+  delegatedToAccount: one(account, {
+    fields: [accountBalance.delegate],
+    references: [account.id],
+    relationName: "delegatedBalances",
   }),
   token: one(token, {
     fields: [accountBalance.tokenId],
@@ -239,9 +243,6 @@ export const accountPowerRelations = relations(
       fields: [accountPower.accountId],
       references: [account.id],
       relationName: "accountPowers",
-    }),
-    delegatedFrom: many(accountBalance, {
-      relationName: "delegatedFromTo",
     }),
   }),
 );
@@ -311,5 +312,8 @@ export const accountRelations = relations(account, ({ many }) => ({
   }),
   votes: many(votesOnchain, {
     relationName: "voter",
+  }),
+  delegatedFromBalances: many(accountBalance, {
+    relationName: "delegatedBalances",
   }),
 }));
--- a/apps/indexer/src/lib/event-handlers.ts
+++ b/apps/indexer/src/lib/event-handlers.ts
@@ -479,7 +479,7 @@ export const voteCast = async (
     proposalId: String(proposalId),
     voterAccountId: event.args.voter,
     support: event.args.support.toString(),
-    weight: weight.toString(),
+    votingPower: weight.toString(),
     reason: event.args.reason,
     timestamp: event.block.timestamp,
   });
--- a/apps/local-node/script/DelegateTokens.s.sol
+++ b/apps/local-node/script/DelegateTokens.s.sol
@@ -36,27 +36,27 @@ contract DelegateTokens is BaseScript, Test {
         // Bob delegates to himself
         console.log("Setting up Bob's delegation...");
         vm.startBroadcast(Constants.BOB_PRIVATE_KEY);
-        ensToken.delegate(Constants.BOB);
+        ensToken.delegate(Constants.CHARLIE);
         vm.stopBroadcast();
         console.log("SUCCESS: Bob delegated to himself");
         
         // Assert Bob now has voting power equal to his token balance
-        assertEq(ensToken.getVotes(Constants.BOB), ensToken.balanceOf(Constants.BOB), "Bob's voting power should equal his token balance after delegation");
+        assertEq(ensToken.getVotes(Constants.CHARLIE), ensToken.balanceOf(Constants.BOB), "Charlies's voting power should equal Bob's token balance after delegation");
         
         // Assert Charlie has token balance before delegation
-        assertGt(ensToken.balanceOf(Constants.CHARLIE), 0, "Charlie must have token balance before delegation");
-        // Assert Charlie has no voting power before delegation
-        assertEq(ensToken.getVotes(Constants.CHARLIE), 0, "Charlie should have no voting power before delegation");
+        assertGt(ensToken.balanceOf(Constants.BOB), 0, "Bob must have token balance before delegation");
+        // Assert Bob has no voting power before delegation
+        assertEq(ensToken.getVotes(Constants.BOB), 0, "Bob should have no voting power before delegation");
         
         // Charlie delegates to himself
         console.log("Setting up Charlie's delegation...");
         vm.startBroadcast(Constants.CHARLIE_PRIVATE_KEY);
-        ensToken.delegate(Constants.CHARLIE);
+        ensToken.delegate(Constants.BOB);
         vm.stopBroadcast();
         console.log("SUCCESS: Charlie delegated to himself");
         
         // Assert Charlie now has voting power equal to his token balance
-        assertEq(ensToken.getVotes(Constants.CHARLIE), ensToken.balanceOf(Constants.CHARLIE), "Charlie's voting power should equal his token balance after delegation");
+        assertEq(ensToken.getVotes(Constants.BOB), ensToken.balanceOf(Constants.CHARLIE), "Bob's voting power should equal Charlie's token balance after delegation");
         
         // Display final voting power
         displayVotingPower("Final");


apps/indexer/src/api/services/historical-balances/historical-balances.service.ts
```diff
@@ -0,0 +1,272 @@
...
+  }: HistoricalBalancesRequest): Promise<HistoricalBalance[]> {
+    const tokenAddress = this.getTokenAddress(daoId);
+
+    if (!tokenAddress) {
+      throw new Error(`Token address not found for DAO: ${daoId}`);
+    }
+
+    try {
+      // Use Multicall3 for blocks 19+ (after Multicall3 deployment), individual calls for earlier blocks
+      if (blockNumber >= 19) {
```
pikonha, line 60: is this something the only happens with ethereum?

apps/indexer/src/api/services/historical-balances/historical-balances.service.ts
```diff
@@ -0,0 +1,272 @@
...
+        return await this.getBalancesWithMulticall(addresses, blockNumber, tokenAddress);
+      } else {
+        return await this.getBalancesIndividually(addresses, blockNumber, tokenAddress);
+      }
```
pikonha, line 64:
```suggestion
```

apps/indexer/src/api/services/historical-balances/historical-balances.service.ts
```diff
@@ -0,0 +1,272 @@
...
+    const contractInfo = CONTRACT_ADDRESSES[network][daoId];
+
+    if (!contractInfo?.token) {
+      throw new Error(
+        `Token contract not found for DAO: ${daoId} on network: ${network}`,
+      );
+    }
+
+    return contractInfo.token;
+  }
```
pikonha, line 248: this should be injected as argument of this service method instead of relying on the env within the inner layers of the application

apps/indexer/src/api/services/historical-balances/historical-balances.service.ts
```diff
@@ -0,0 +1,272 @@
...
+            { name: "returnData", type: "bytes[]" },
+          ],
+          stateMutability: "payable",
+          type: "function",
+        },
+      ] as const,
+      functionName: "aggregate",
+      args: [calls],
+      blockNumber: BigInt(blockNumber),
+    });
```
pikonha, line 130: why no to use the viem's [multicall](https://viem.sh/docs/contract/multicall.html) method? it has the [address of the contract on each chain](https://github.com/wevm/viem/blob/main/src/chains/definitions/arbitrum.ts#L20) (except anvil)

apps/indexer/src/api/services/historical-balances/historical-balances.service.ts
```diff
@@ -0,0 +1,272 @@
...
+        ] as const,
+        functionName: "balanceOf",
+        args: [address],
+      });
+
+      return {
+        target: tokenAddress,
+        callData: calldata,
+      };
+    });
```
pikonha, line 101:
```suggestion
    const calls = addresses.map(address => ({
      encodeFunctionData({
      abi: parseAbi(["function balanceOf(address account) (uint256)"]),
        functionName: "balanceOf",
        args: [address],
      });
     })
   )
```

apps/indexer/src/api/services/historical-balances/historical-balances.service.ts
```diff
@@ -0,0 +1,272 @@
...
+}
+
+export interface HistoricalBalancesRequest {
+  addresses: Address[];
+  blockNumber: number;
+  daoId: DaoIdEnum;
+}
+
+// Multicall3 contract address (deployed after governance setup)
+const MULTICALL3_ADDRESS = "0xA51c1fc2f0D1a1b8494Ed1FE312d7C3a78Ed91C0" as Address;
```
pikonha, line 23: this already exists within [viem's chain config](https://github.com/wevm/viem/blob/main/src/chains/definitions/arbitrum.ts#L20)

apps/indexer/src/api/services/historical-balances/historical-balances.service.ts
```diff
@@ -0,0 +1,272 @@
...
+
+    const balances = await Promise.all(balancePromises);
+
+    // Transform results into HistoricalBalance objects
+    return addresses.map((address, index) => ({
+      address: address!,
+      balance: balances[index] || 0n,
+      blockNumber,
+      tokenAddress,
+    }));
```
pikonha, line 186:
```suggestion
const balances = await Promise.allSettled(
      addresses.map((address) =>
        this.client.readContract({
          address: tokenAddress,
          abi: parseAbi(["function balanceOf(address account) (uint256)"]),
          functionName: "balanceOf",
          args: [address],
          blockNumber: BigInt(blockNumber),
        }),
      ),
    );

    // Transform results into HistoricalBalance objects
    return addresses.map((address, index) => ({
      address,
      balance:
        balances[index]?.status === "fulfilled"
          ? (balances[index].value as bigint)
          : 0n,
      blockNumber,
      tokenAddress,
    }));
```

apps/indexer/src/api/services/historical-balances/historical-balances.service.ts
```diff
@@ -0,0 +1,272 @@
...
+    const chain = getChain(chainId);
+
+    if (!chain) {
+      throw new Error(`Chain with ID ${chainId} not found`);
+    }
+
+    this.client = createPublicClient({
+      chain,
+      transport: http(rpcUrl),
+    });
```
pikonha, line 39: all of these could be handled on the controller and returning 500 if it fails

apps/indexer/src/api/services/historical-balances/README.md
pikonha, line 1: this is pretty self explanatory with the [route definition](https://github.com/blockful-io/anticapture/pull/871/files#diff-630afb629e5d9d2665e4396df3ac1d9c7910d13a21c72864574396dc06b2ae8aR15) imho

apps/indexer/src/api/controller/historical-balances.controller.ts
```diff
@@ -0,0 +1,209 @@
...
+      description:
+        "Fetch historical token balances for multiple addresses at a specific block number using multicall",
+      tags: ["historical-balances"],
+      request: {
+        params: z.object({
+          daoId: caseInsensitiveEnum(DaoIdEnum),
+        }),
+        query: z.object({
+          addresses: z.string().min(1, "At least one address is required"),
+          blockNumber: z.coerce.number().int().positive("Block number must be positive"),
```
pikonha, line 30: do we really need a positive verification on this?

pikonha (changes requested):
a lot of complexity that could be handled by the already established route validation by zod. it would be good to have the endpoint and its response clear on the PR description

apps/indexer/src/api/controller/historical-balances.controller.ts
```diff
@@ -0,0 +1,209 @@
...
+            balance: balance.balance.toString(), // Convert BigInt to string for JSON serialization
+            blockNumber: balance.blockNumber,
+            tokenAddress: balance.tokenAddress,
+          })),
+          metadata: {
+            totalAddresses: addresses.length,
+            blockNumber,
+            daoId,
+            tokenAddress,
+          },
```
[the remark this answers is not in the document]
lgahdl: I'll remove it, thanks for the suggestion

apps/local-node/script/Constants.sol
```diff
@@ -21,9 +21,11 @@ library Constants {
...
     
     // Contract addresses (deterministic with Anvil)
-    address internal constant ENS_TOKEN_ADDRESS = 0x5FbDB2315678afecb367f032d93F642f64180aa3;
-    address internal constant ENS_GOVERNOR_ADDRESS = 0x9fE46736679d2D9a65F0992F2272dE9f3c7fa6e0;
-    address internal constant ENS_TIMELOCK_ADDRESS = 0xe7f1725E7734CE288F8367e1Bb143E90bb3F0512;
+    // Updated to match actual deployment addresses from latest run
+    address internal constant ENS_TOKEN_ADDRESS = 0x5FbDB2315678afecb367f032d93F642f64180aa3;    // ✅ Correct
+    address internal constant ENS_GOVERNOR_ADDRESS = 0x9fE46736679d2D9a65F0992F2272dE9f3c7fa6e0;  // ✅ Updated
+    address internal constant ENS_TIMELOCK_ADDRESS = 0xe7f1725E7734CE288F8367e1Bb143E90bb3F0512;   // ✅ Updated  
+    address internal constant MULTICALL3_ADDRESS = 0xA51c1fc2f0D1a1b8494Ed1FE312d7C3a78Ed91C0;    // ✅ Fixed - actual deployed address
```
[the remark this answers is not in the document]
lgahdl: I deployed the Multicall3 in the local node, but I will go to the approach with the fork and remove this

apps/indexer/src/api/controller/historical-balances.controller.ts
lgahdl: Why we would allow negative numbers?

apps/indexer/package.json
```diff
@@ -3,7 +3,7 @@
   "version": "0.3.1",
   "private": true,
   "scripts": {
-    "dev": "ponder dev --schema public",
+    "dev": "ponder dev --schema public --config config/ens.local.config.ts",
```
pikonha, line 6: I usually use the dev command with the ens config file, amd this change is enforcing me to use the local config

apps/indexer/src/api/controller/historical-balances.controller.ts
```diff
@@ -0,0 +1,103 @@
...
+          description: "Internal server error",
+          content: {
+            "application/json": {
+              schema: z.object({
+                error: z.string(),
+                message: z.string(),
+              }),
+            },
+          },
+        },
```
pikonha, line 81: if you add this the api gateway will require the `...200Reponse` fragment on every fetch. therefore, I'd remove this for now

```suggestion
```

apps/indexer/src/api/controller/historical-balances.controller.ts
```diff
@@ -0,0 +1,103 @@
...
+        params: z.object({
+          daoId: caseInsensitiveEnum(DaoIdEnum),
+        }),
+        query: z.object({
+          addresses: z
+            .string()
+            .min(1, "At least one address is required")
+            .transform((str) =>
+              str.split(",").map((addr) => addr.trim() as Address)
+            ),
```
pikonha, line 34: zod has an [array validation builtin](https://zod.dev/api?id=arrays)

```suggestion
          addresses: z.array(z.string())
            .min(1, "At least one address is required"),
```

apps/indexer/src/api/controller/historical-balances.controller.ts
```diff
@@ -0,0 +1,103 @@
...
+            .positive("Block number must be positive"),
+        }),
+      },
+      responses: {
+        200: {
+          description: "Successfully retrieved historical balances",
+          content: {
+            "application/json": {
+              schema: z.object({
+                data: z.array(
```
pikonha, line 47: this `data` object isn't really needed, its only adding a layer on the graphql query

apps/indexer/src/api/controller/historical-balances.controller.ts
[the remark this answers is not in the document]
pikonha: double down on this one

apps/indexer/src/api/services/historical-balances/historical-balances.service.ts
```diff
@@ -0,0 +1,163 @@
...
+    addresses,
+    blockNumber,
+    daoId,
+  }: HistoricalBalancesRequest): Promise<HistoricalBalance[]> {
+    const tokenAddress = this.getTokenAddress(daoId);
+
+    if (!tokenAddress) {
+      throw new Error(`Token address not found for DAO: ${daoId}`);
+    }
+
```
pikonha, line 61: how is this being handled by the API?

apps/indexer/src/api/services/historical-balances/historical-balances.service.ts
```diff
@@ -0,0 +1,163 @@
...
+        functionName: "balanceOf",
+        args: [address],
+      })),
+      blockNumber: BigInt(blockNumber),
+    });
+
+    // Transform results into HistoricalBalance objects
+    return addresses.map((address, index) => {
+      const result = results[index];
+      const balance = result?.status === "success" ? (result.result ?? 0n) : 0n;
```
pikonha, line 102:
```suggestion
      const balance = result?.result|| 0n;
```

apps/indexer/src/api/services/historical-balances/historical-balances.service.ts
lgahdl: The input of the API gets only the daoId(UNI, ENS, OP), and then get the token address from the CONTRACT_ADDRESSES, so this line will never be accessed, since the wrong daoId triggers a validation error:
`{"error":"Validation Error","message":"Validation error: Expected 'UNI' | 'ENS' | 'ARB', received null at \"daoId\"","details":[{"expected":"'UNI' | 'ENS' | 'ARB'","received":"null","code":"invalid_type","path":["daoId"],"message":"Expected 'UNI' | 'ENS' | 'ARB', received null"}]}`

apps/indexer/src/api/controller/historical-balances.controller.ts
pikonha: please remove it

pikonha (changes requested):

PR #962, opened 2025-06-18
fix: use pnpm on the dashboard

pikonha:


lgahdl (approved):
The storybook deployment is broken, but overall looks good.

PR #975, opened 2025-06-25
Logic from DAO config

pikonha:
Uni has logic as `For`, but it's being shown `For + Abstain`. This PR then moves this logic to the DAO config structure so each DAO has its own.


lgahdl (approved):
Looks good!

PR #998, opened 2025-07-02
fix: Calculating avgTimeBeforeEnd (was undone);

lgahdl:
fix: Adding view schema name before the queries;

This PR is to fix 2 bugs in the Proposals Activity HTTP Endpoint:
- Serialization error for bigints in the forVotes, abstainVotes and againstVotes;
- avgTimeBeforeEnd was not being calculated

### DIFF
--- a/apps/indexer/src/api/controller/proposals-activity.controller.ts
+++ b/apps/indexer/src/api/controller/proposals-activity.controller.ts
@@ -73,9 +73,9 @@ export function proposalsActivity(
                       endBlock: z.string(),
                       timestamp: z.string(),
                       status: z.string(),
-                      forVotes: z.number(), // default 0
-                      againstVotes: z.number(), // default 0
-                      abstainVotes: z.number(), // default 0
+                      forVotes: z.string(),
+                      againstVotes: z.string(),
+                      abstainVotes: z.string(),
                     }),
                     userVote: z
                       .object({
--- a/apps/indexer/src/api/repositories/drizzle.ts
+++ b/apps/indexer/src/api/repositories/drizzle.ts
@@ -14,14 +14,14 @@ export class DrizzleRepository {
     const query = sql`
       WITH old_data AS (
         SELECT db.average as old_amount
-        FROM dao_metrics_day_buckets db
+        FROM anticapture.dao_metrics_day_buckets db
         WHERE db."metricType" = ${metricType}
         AND db."date" <= ${this.now() - days}
         ORDER BY db."date" desc LIMIT 1
       ),
       current_data AS (
         SELECT db.average as current_amount
-        FROM dao_metrics_day_buckets db
+        FROM anticapture.dao_metrics_day_buckets db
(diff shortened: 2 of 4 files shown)


apps/indexer/src/api/services/proposals-activity/proposals-activity.service.ts
```diff
@@ -227,6 +235,40 @@ export class ProposalsActivityService {
     return statusUpper === "EXECUTED" ? "1" : "0"; // "1" = For, "0" = Against
   }
```
edulennert, line 237: Would add some comments here, for example, the return type is in seconds?

edulennert (approved):

apps/indexer/src/api/controller/proposals-activity.controller.ts
```diff
@@ -73,9 +73,9 @@ export function proposalsActivity(
                       endBlock: z.string(),
                       timestamp: z.string(),
                       status: z.string(),
-                      forVotes: z.number(), // default 0
-                      againstVotes: z.number(), // default 0
-                      abstainVotes: z.number(), // default 0
+                      forVotes: z.string(),
+                      againstVotes: z.string(),
+                      abstainVotes: z.string(),
```
LeonardoVieira1630, line 78: why is it a string right now?

apps/indexer/src/api/repositories/drizzle.ts
```diff
@@ -14,14 +14,14 @@ export class DrizzleRepository {
     const query = sql`
       WITH old_data AS (
         SELECT db.average as old_amount
-        FROM dao_metrics_day_buckets db
+        FROM anticapture.dao_metrics_day_buckets db
```
pikonha, line 17: this shouldn't be added, the `view` created by ponder is a [materialized view](https://www.postgresql.org/docs/current/rules-materializedviews.html) that works the same way as a schema

apps/indexer/src/api/controller/proposals-activity.controller.ts
pikonha: I don't see the reason either

pikonha (changes requested):
please explain why this PR is needed

apps/indexer/src/api/controller/proposals-activity.controller.ts
lgahdl: Serialization problems, too big numbers become 1.2345e+22, and we get an unexpected response type error

lgahdl: All these numbers comes as the voting power with 18 decimals, that's why they're big numbers btw

lgahdl:
> please explain why this PR is needed

Just added the explanation in the description

apps/indexer/src/api/services/proposals-activity/proposals-activity.service.ts
pikonha: same

PR #1009, opened 2025-07-04
fix: use railway deployment id as indexer schema

pikonha:
Railway doesn't seem to pull the `.git` directory on each deploy

image

We need to have a unique schema for each deployment so the Ponder don't acuses a schema conflict.

image


LeonardoVieira1630 (approved):
Sad to approve this. Kinda liked the old solution.

PR #1036, opened 2025-07-09
Refactor/controllers endpoints to remove dao id parameter

LeonardoVieira1630:
## Summary

  Refactor API controllers to remove unnecessary daoId parameters from endpoint functions where the DAO context is already available through environment configuration.

 ## Changes

  - Removed daoId parameter from controllers that don't need it:
    - assets.controller.ts - Uses external Dune API, DAO context not required
    - governance-activity/controller.ts - DAO context available via environment
    - token-distribution/controller.ts - DAO context available via environment
  - Updated controller instantiation in api/index.ts to match new signatures
  - Maintained daoId parameter for controllers that still need it:
    - proposals-activity.controller.ts - Passes daoId to service layer
    - historical-onchain.controller.ts - Requires DAO-specific contract
  addresses

 ## Rationale

  Since each indexer instance serves a single DAO (configured via env.DAO_ID), most controllers don't need the daoId as an explicit parameter. This simplifies the API surface while maintaining the same functionality.

 ## Future Improvements

  Consider removing daoId fields from the database schema and service layers entirely, since each indexer instance is dedicated to a single DAO. This would further simplify the codebase by eliminating the need for DAO-specific filtering in queries and reducing complexity across controllers and services.


apps/indexer/src/api/controller/token-historical-data.controller.ts
```diff
@@ -12,27 +11,24 @@ import { DAYS_IN_YEAR } from "@/lib/constants";
...
+      path: "/token/historical-data",
       summary: "Get historical token data",
       description: "Get historical market data for a specific token",
       tags: ["tokens"],
-      request: {
-        params: z.object({
-          daoId: caseInsensitiveEnum(DaoIdEnum),
-        }),
-      },
+      request: {},
```
pikonha, line 31:
```suggestion
```

apps/indexer/src/api/controller/token-historical-data.controller.ts
```diff
@@ -1,7 +1,6 @@
 import { OpenAPIHono as Hono, createRoute, z } from "@hono/zod-openapi";
 
 import { DaoIdEnum } from "@/lib/enums";
-import { caseInsensitiveEnum } from "../middlewares";
```
pikonha, line 4: you can probably remove this middleware

pikonha (approved):
on this PR, I think you could go through the `graphql-client/documents/*.graphql` files, removing the daoID and whatever it is used like so `daoId={daoId as unknown as QueryInput_HistoricalVotingPower_DaoId}`

PR #1038, opened 2025-07-09
fix: Fixing attack profitability chart timestamp conversion and using…

lgahdl:
… treasury "close" instead of "high";

### DIFF
--- a/apps/dashboard/features/attack-profitability/utils/normalizeDatasetAllTreasury.ts
+++ b/apps/dashboard/features/attack-profitability/utils/normalizeDatasetAllTreasury.ts
@@ -27,7 +27,7 @@ export function normalizeDatasetAllTreasury(
   const sortedGovernanceTokenTreasuries = [...(governanceTokenTreasuries ?? [])]
     .map((item) => ({
       timestamp: Number(item.date) * 1000,
-      close: Number(item.close) / 1e18,
+      close: Number(item.close) / 1e18, // close is the price in the end of the day, for treasury is better if we use this instead of high because generally speaking there is not too many changes in the treasury during the year
     }))
     .sort((a, b) => a.timestamp - b.timestamp);
 


heronlancellot (approved):

apps/dashboard/features/attack-profitability/utils/normalizeDataset.ts
```diff
@@ -25,7 +25,7 @@ export function normalizeDataset(
   // Prepare multipliers sorted by timestamp
   const sortedMultipliers = multiplierDataSet
     .map((item) => ({
-      timestamp: Number(item.date),
+      timestamp: Number(item.date) * 1000,
```
edulennert, line 28: magic numbers?

apps/dashboard/features/attack-profitability/utils/normalizeDatasetAllTreasury.ts
```diff
@@ -26,8 +26,8 @@ export function normalizeDatasetAllTreasury(
 
   const sortedGovernanceTokenTreasuries = [...(governanceTokenTreasuries ?? [])]
     .map((item) => ({
-      timestamp: Number(item.date),
-      high: Number(item.high) / 1e18,
+      timestamp: Number(item.date) * 1000,
+      close: Number(item.close) / 1e18,
```
edulennert, line 30: would appreciate some comments on what close means

edulennert (approved):

pikonha (changes requested):
the chart doesn't seem to reflect the delegation change on preview

image

apps/dashboard/features/attack-profitability/utils/normalizeDataset.ts
lgahdl: no, the timestamp is in seconds, I need to convert to miliseconsd, that's all hahahahahahaha

apps/dashboard/features/attack-profitability/utils/normalizeDatasetAllTreasury.ts
lgahdl: Is the value for treasury in the end of the day, it's more suitable for this case, because since treasury doesn't have too many transactions changing it, we will only see the treasury decaying from a transaction made in march after other transaction is made in April, in the chart looks like the treasury dacayed in April, but actually decayed in March. I will add comments for this, but just explaining why I changed it

PR #1040, opened 2025-07-10
Feat/delegate voting power

### LINKED ISSUE #932: feat: Delegate Voting Power
## Feature Request

Image

```graphql
query MyQuery {
  accountBalances(
    orderBy: "balance"
    orderDirection: "desc"
    limit: 10
    where: {delegate: "0x5BFCB4BE4d7B43437d5A0c57E908c048a4418390"}
  ) {
    items {
      delegate
      balance
    }
    pageInfo {
      hasNextPage
      hasPreviousPage
      startCursor
      endCursor
    }
  }
}
```

heronlancellot:
Closes #932


apps/dashboard/features/holders-and-delegates/components/Delegates.tsx
```diff
@@ -246,11 +247,7 @@ export const Delegates = ({
           onClick={() => handleSort("votingPower")}
         >
           <h4 className="text-table-header">
-            Voting Power (
-            {daoId === QueryInput_HistoricalVotingPower_DaoId.Ens
-              ? "ENS"
-              : "UNI"}
-            )
+            Voting Power ({daoId === DaoIdEnum.ENS ? "ENS" : "UNI"})
```
pikonha, line 250: we can't rely on just these two daos

apps/dashboard/features/holders-and-delegates/components/ThePieChart.tsx
```diff
@@ -0,0 +1,189 @@
...
+        </p>
+        <p className="text-xs text-gray-600">
+          Percentage: <strong>{percentage}%</strong>
+        </p>
+      </div>
+    </div>
+  );
+};
+
+const renderCustomizedLabel = ({
```
lgahdl, line 77: I'd rather have this function in the utils, instead of here, this is a utilitary function for this feature.

lgahdl (changes requested):
A lot of repeated stuff, you should probably change the base branch of this one

isadorable-png (commented):
- Show ENS domains (right now only showing the wallet address)
- Change tooltip color (use our dark theme palette)
- Can we have a hover on the pie chart slice?
- Change the table label row height to 32px
-

apps/dashboard/shared/hooks/graphql-client/useVotingPower.ts
```diff
@@ -0,0 +1,342 @@
...
+
+  // ------------------------------------------------------------------
+  // Enrich balances with timestamp (fallback to undefined)
+  // ------------------------------------------------------------------
+  const balancesWithTimestamp = (accountBalances || []).map((account) => ({
+    ...account,
+    timestamp: timestampMap[account.accountId.toLowerCase()],
+  }));
+
+  const { data: top5Delegators } = useGetTop5DelegatorsQuery({
```
lgahdl, line 172: I'm not a big fan of having a number here, can't you change to "useGetTopFiveDelegatorsQuery" and "topFiveDelegators" or just "useGetTopDelegatorsQuery" and "topDelegators"?

apps/dashboard/shared/hooks/graphql-client/useVotingPower.ts
```diff
@@ -0,0 +1,342 @@
...
+  historicalDataLoading: boolean;
+}
+interface UseVotingPowerParams {
+  daoId: DaoIdEnum;
+  address: string;
+  orderBy?: string;
+  orderDirection?: string;
+}
+
+export const useVotingPower = ({
```
lgahdl, line 58: this hook is huge, I think you should split it into more files, if possible.

edulennert (changes requested):
Shouldn't be possible to have both the filters active
image

There is a glitch in the loading state of the table
image

edulennert (approved):
Feel free to merge after those 2 things and QA Design review. Code looks good 🙂

apps/dashboard/shared/hooks/graphql-client/useVotingPower.ts
heronlancellot: The problem of this hook is not the implementation of it, but the Pagination logic inside of it..

isadorable-png (commented):
Looking awesome!! Just some more small details:
image
- Table sorting is not working correctly, it's not sorting the whole data, just from the current page.

---

image

- Can we increase a bit the percentage value that we're going to display the label? The text is overflowing is some cases still.
- The hover on the chart slice is looking awesome, but I'd like to increase its contrast. Can we decrease a bit more the opacity?

---

https://github.com/user-attachments/assets/cac5d15f-3e78-4d2e-b481-969857ce7742


- While loading we need to display a skeleton, right now the blank slate is being shown.

isadorable-png (approved):
Looking great!

PR #1059, opened 2025-07-15
fix: remove daoId from rest and use dayEnum instead of blocknumber

pikonha:


apps/indexer/config/ens.local.config.ts
```diff
@@ -1,23 +1,19 @@
...
 export default createConfig({
-  database: {
-    kind: "postgres",
-    connectionString: env.DATABASE_URL,
-  },
   chains: {
     anvil: {
       id: 31337,
-      rpc: env.RPC_URL,
+      rpc: "http://localhost:8545",
```
heronlancellot, line 13: hey yo, Why it's not in .env ?

heronlancellot (commented):
Hey yo, something crash in delegates
image

lgahdl (approved):

PR #1165, opened 2025-08-06
sync prod with dev

pikonha:


apps/indexer/src/eventHandlers/delegation.ts
```diff
@@ -47,16 +47,25 @@ export const delegateChanged = async (
...
-    timestamp,
-    logIndex,
-  });
+  if (!delegatorBalance) {
+    throw new Error(`Delegator ${delegator} not found`);
+  }
+
+  await context.db
+    .insert(delegation)
+    .values({
```
pikonha, line 56: The onConflictDoUpdate method is missing the target columns specification. This will cause a runtime error because the database doesn't know which columns constitute a conflict. You need to specify the target columns like: .onConflictDoUpdate({ target: [delegation.transactionHash, delegation.logIndex], set: { delegatedValue: delegatorBalance.balance } })

pikonha (commented):
Automated code review

PR #1184, opened 2025-08-08
feat: completed last update endpoint, returns the last update date fo…

lgahdl:
…r each chart;

### DIFF
--- a/apps/indexer/src/env.ts
+++ b/apps/indexer/src/env.ts
@@ -6,7 +6,7 @@ dotenv.config();
 
 const envSchema = z.object({
   RPC_URL: z.string(),
-  DATABASE_URL: z.string().optional().default(""),
+  DATABASE_URL: z.string().optional(),
   POLLING_INTERVAL: z.coerce.number().default(10000), // 10s
   MAX_REQUESTS_PER_SECOND: z.coerce.number().default(20),
   DAO_ID: z.nativeEnum(DaoIdEnum),


apps/indexer/src/api/controller/last-update.controller.ts
```diff
@@ -0,0 +1,45 @@
...
+    }),
+    async (context) => {
+      const { chart } = context.req.valid("query");
+
+      try {
+        const lastUpdate = await service.getLastUpdate(chart);
+        return context.json({ lastUpdate }, 200);
+      } catch (error) {
+        console.error("Error getting last update:", error);
+        return context.json({ lastUpdate: null }, 200);
```
pikonha, line 41: error as 200?

apps/indexer/src/api/controller/last-update.controller.ts
```diff
@@ -0,0 +1,45 @@
...
+          chart: z.nativeEnum(ChartType),
+        }),
+      },
+      responses: {
+        200: {
+          description: "Last update time",
+          content: {
+            "application/json": {
+              schema: z.object({
+                lastUpdate: z.string().nullable(),
```
pikonha, line 26: when will this be null?

apps/indexer/src/api/controller/last-update.controller.ts
pikonha: shouldn't that be a 404?

apps/indexer/src/api/repositories/last-update.repository.ts
```diff
@@ -0,0 +1,53 @@
...
+    );
+
+    const whereCondition =
+      metricConditions.length === 1
+        ? metricConditions[0]
+        : or(...metricConditions);
+
+    // Find the record with the greatest timestamp for the specified metrics
+    const lastUpdate = await db.query.daoMetricsDayBucket.findFirst({
+      where: whereCondition,
```
pikonha, line 47: you could use [drizzle's inArray](https://orm.drizzle.team/docs/operators#inarray) instead

apps/indexer/src/api/repositories/last-update.repository.ts
```diff
@@ -0,0 +1,53 @@
...
+      case ChartType.TokenDistribution:
+        metricsToCheck = [
+          MetricTypesEnum.DELEGATED_SUPPLY,
+          MetricTypesEnum.CEX_SUPPLY,
+          MetricTypesEnum.DEX_SUPPLY,
+          MetricTypesEnum.LENDING_SUPPLY,
+        ];
+        break;
+      default:
+        throw new Error(`Unsupported chart type: ${chart}`);
```
pikonha, line 32: couldn't this simply return null instead of throwin an exception?

pikonha (changes requested):
it would be good to add the label on the charts as well so it is fully integrated

apps/indexer/src/api/controller/last-update.controller.ts
lgahdl: error 404 returns lastUpdate as null

apps/indexer/src/api/repositories/last-update.repository.ts
```diff
@@ -0,0 +1,42 @@
...
+        ];
+        break;
+      default:
+        break;
+    }
+    // Find the record with the greatest timestamp for the specified metrics
+    const lastUpdate = await db.query.daoMetricsDayBucket.findFirst({
+      where: inArray(daoMetricsDayBucket.metricType, metricsToCheck),
+      orderBy: (fields, { desc }) => [desc(fields.date)],
+    });
```
pikonha, line 38:
```suggestion
    const lastUpdate = await db.query.daoMetricsDayBucket.findFirst({
      columns: {
        date: true,
      },
      where: inArray(daoMetricsDayBucket.metricType, metricsToCheck),
      orderBy: (fields, { desc }) => [desc(fields.date)],
    });
```

apps/indexer/src/api/repositories/last-update.repository.ts
```diff
@@ -0,0 +1,42 @@
...
+
+    return lastUpdate?.date ?? null;
```
pikonha, line 43:
```suggestion
    return lastUpdate?.date;
```

pikonha (changes requested):
theres a typescript type mismatch 

Image

apps/indexer/src/api/repositories/last-update.repository.ts
```diff
@@ -0,0 +1,45 @@
...
+        break;
+    }
+    // Find the record with the greatest timestamp for the specified metrics
+    const lastUpdate = await db.query.daoMetricsDayBucket.findFirst({
+      columns: {
+        date: true,
+      },
+      where: inArray(daoMetricsDayBucket.metricType, metricsToCheck),
+      orderBy: (fields, { desc }) => [desc(fields.date)],
+    });
```
pikonha, line 41: This won't work its purpose since we reset the time of each daily metric to midnight

Image

pikonha (changes requested):

pikonha (changes requested):
the `delegations.ts` file is truncating the date before calling the `storeDailyBucket`. it should be removed from there

pikonha (approved):

PR #1215, opened 2025-08-13
Prod deploy

pikonha:


apps/api-gateway/schema.graphql
```diff
@@ -0,0 +1,1451 @@
+directive @enum(subgraph: String, value: String) on ENUM_VALUE
```
pikonha, line 1: The schema.graphql file is added but was previously in .gitignore. This could break builds if the file is expected to be generated rather than committed. Verify if this file should be generated during build or committed to version control.

.github/workflows/deploy-indexer.yaml
```diff
@@ -77,6 +77,17 @@ jobs:
...
       - uses: actions/checkout@v3
+      - name: Install Node.js and pnpm
+        run: |
+          apk add --no-cache nodejs npm
+          npm install -g pnpm@10.10.0
+      - name: Install dependencies
+        run: pnpm install --frozen-lockfile
+      - name: Prune DB inactive schemas
+        if: ${{ needs.changes.outputs.indexer == 'true' }}
+        run: |
```
pikonha, line 88: Database pruning operations in CI/CD without proper error handling could cause deployment failures. The db:prune command should include error handling and rollback mechanisms in case of failures.

apps/dashboard/features/holders-and-delegates/hooks/useDelegates.ts
```diff
@@ -373,14 +380,21 @@ export const useDelegates = ({
...
+    refetch({
+      after: undefined,
+      before: undefined,
+      orderBy,
+      orderDirection,
+      ...(address && { addresses: [address] }),
+    });
+  }, [refetch, orderBy, orderDirection, address]);
 
   return {
```
pikonha, line 399: Missing error handling for the address filter parameter. If an invalid address format is passed, it could cause GraphQL query errors. Add validation for the address parameter format before including it in the query.

pikonha (commented):
Automated code review

Issue #729, opened 2025-05-08
fix: Sidebar items shouldn't be active when autoscrolling to another tab


lgahdl:
## Bug Report

https://github.com/user-attachments/assets/04d5c969-0858-44da-afae-2a39ea7c5144

PR #1229, opened 2025-08-15
Fix/Sidebar active indicator sliding throw all the items

Legacynnn:
### **Summary** 
This PR fixes the sidebar active indicator that was sliding through all intermediate items when navigating between one item to another if the user clicks on it. The indicator now transitions directly to the selected item, keeping the on scroll change.

### **Changes**
Context

- `activeSection: string | null` state to track currently active sidebar section
- `updateActiveSection` method with options parameter for granular control:
- `source: "programmatic" | "event"` for distinguishes between user clicks and scroll events
- `immediate: boolean` forces immediate state updates when needed
- `end: boolean` signals end of programmatic navigation

### **Issue**
https://github.com/blockful/anticapture/issues/729#issue-3049044501

New behavior

https://github.com/user-attachments/assets/ee2b8c43-d303-4cf0-a00d-ac5dd5638c9b

Disclaimer: I chose this approach trying to avoid overuse props on the context, but I'll be very happy to reimplement it differently if that would better suit the project's conventions! Awesome project btw 🫡

### DIFF
--- a/apps/dashboard/shared/hooks/useSectionObserver.tsx
+++ b/apps/dashboard/shared/hooks/useSectionObserver.tsx
@@ -1,6 +1,7 @@
 "use client";
 
-import { useEffect, useState, useRef } from "react";
+import { useEffect, useRef } from "react";
+import { useDaoPageInteraction } from "@/shared/contexts/DaoPageInteractionContext";
 
 interface UseSectionObserverProps {
   initialSection?: string;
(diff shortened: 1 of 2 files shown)


apps/dashboard/shared/hooks/useSectionObserver.tsx
```diff
@@ -9,28 +10,23 @@ interface UseSectionObserverProps {
...
     const handleSectionChange = (event: Event) => {
-      if (isScrollingRef.current) return;
       const customEvent = event as CustomEvent<string>;
-      setActiveSection(customEvent.detail);
+      updateActiveSection(customEvent.detail, { source: "event" });
     };
 
     window.addEventListener("sectionInView", handleSectionChange);
     return () => {
       window.removeEventListener("sectionInView", handleSectionChange);
```
pikonha, line 27: Potential closure issue: `activeSection` in the setTimeout may reference a stale value from when the timeout was created, not the current active section when it executes after 1000ms. Consider capturing the current section ID or removing this parameter since it's updating programmatic state end.

pikonha (commented):
Automated code review

PR #1230, opened 2025-08-15
fix: sidebar active indicator sliding throw all the items

Legacynnn:
### **Summary** 
This PR fixes the sidebar active indicator that was sliding through all intermediate items when navigating between one item to another if the user clicks on it. The indicator now transitions directly to the selected item, keeping the on scroll change.

### **Changes**
Context

- `activeSection: string | null` state to track currently active sidebar section
- `updateActiveSection` method with options parameter for granular control:
- `source: "programmatic" | "event"` for distinguishes between user clicks and scroll events
- `immediate: boolean` forces immediate state updates when needed
- `end: boolean` signals end of programmatic navigation

### **Issue**
https://github.com/blockful/anticapture/issues/729#issue-3049044501

New behavior

https://github.com/user-attachments/assets/ee2b8c43-d303-4cf0-a00d-ac5dd5638c9b

Disclaimer: I chose this approach trying to avoid overuse props on the context, but I'll be very happy to reimplement it differently if that would better suit the project's conventions! Awesome project btw 🫡

### DIFF
--- a/apps/dashboard/shared/contexts/DaoPageInteractionContext.tsx
+++ b/apps/dashboard/shared/contexts/DaoPageInteractionContext.tsx
@@ -1,18 +1,29 @@
 "use client";
 
-import { createContext, useState, useContext, ReactNode } from "react";
+import { createContext, useState, useContext, ReactNode, useRef } from "react";
 import { RiskAreaEnum } from "@/shared/types/enums/RiskArea";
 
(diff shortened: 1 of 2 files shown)


pikonha (changes requested):
There's a bug where it skips the Attack Profitability section for some reason.

Issue #483, opened 2025-03-27
feat: Increment version of packages automatically


lgahdl:
## Feature Request
This feature is to increment the version from package.json automatically, instead of manually

PR #1286, opened 2025-09-04
fix: version to 0.5

heronlancellot:
- [x] fix: version to 0.5


heronlancellot:
hey yo bro @brunod-e , Yess, I totally agree with you, we need to add this functionality to automatically update sonn
https://github.com/blockful/anticapture/issues/483 we have an open issue about this

PR #1287, opened 2025-09-04
Feat/order by voting power

### LINKED ISSUE #1191: feat:  Order by voting power change by amount
# Feature request

We need to have a way to filter voting power change by amount and filter by addresses

- [ ] Add addressFilter into delegator and delegate
- [ ] Create and add AmountFilter
- [ ] adjust the addressFilter with the correct query
- [ ] add this also in delegator / delegate

Image

Amount Filter

Image

heronlancellot:
Closes #1191


pikonha (changes requested):
1. I'm unable to change the min and max value after typing something

Image

2. filtering by >1000000 causes an error

Image

3. when I filter by min == 1, it changes the list, even though it should look the same

Image

PR #1314, opened 2025-09-10
Feat/1239 feat proposal overview

edulennert:


apps/dashboard/features/governance/types/index.ts
```diff
@@ -0,0 +1,41 @@
...
+  EXECUTED = "executed",
+  DEFEATED = "defeated",
+  CANCELLED = "cancelled",
+}
+
+export enum ProposalState {
+  WAITING_TO_START = "waiting_to_start",
+  ACTIVE = "active",
+  COMPLETED = "completed",
+}
```
pikonha, line 13: what's the difference between this 2? the governor only has status

isadorable-png (commented):
image

- missing top navigation


image

- remove "proposal" from the badge label
- change avatar size to 16px (add this size to the DS)
- this top section does not have border or padding
- missing "by" next to user name
- increase title to h4

image

- "Current Results"  and "Quorum" is not roboto mono 
- increase gap between bars
- when the quorum is reached we need the green badge (check Figma)
- when the voting period closed, the orange bar can be hidden
- quorum is hoverable and has a tooltip

image

- font should be regular here
- completed steps should be white not green

pikonha:
image

https://anticapture-git-feat-1239-feat-proposal-overview-ful.vercel.app/ens/governance/proposal/43003012871521685709452092060115706577030630127946926655190871046443785691900

Issue #848, opened 2025-05-27
feat: Table Component Design System


lgahdl:
### User story

As a **developer using the table component**, I want to have a **table that matches the design system specifications and includes advanced features like fixed columns and enhanced pagination** so that **I can create consistent, accessible, and user-friendly data displays across the application**.

As a **user interacting with data tables**, I want to have **better navigation and viewing capabilities** so that **I can efficiently browse large datasets and maintain context while scrolling**.

### Describe the feature request

The current `TheTable` component needs to be adapted to match the Design System specifications provided by the design team. Additionally, we need to implement two new critical features:

1. **Fixed first column with horizontal scroll control** - to maintain context when viewing wide tables
2. **Enhanced pagination system** - to improve navigation through large datasets

The feature request is related to the problem that our current table implementation doesn't follow the established design system, lacks important UX features, and doesn't provide the level of functionality needed for complex data displays.

##### **1. Label/Column Component Variants**
Based on the design specifications, implement the following variants:

![Image](https://github.com/user-attachments/assets/1e92ca5c-edc8-4a44-bad8-be2c68ebe9c8)

**Size Variants:**
- `DEFAULT` - Standard height and spacing
- `SMALL` - Compact version with reduced padding

**State Variants:**
- `DEFAULT` - Normal state
- `HAS_SORT` - With sort indicators (↑↓)
- `HAS_FILTER` - With filter dropdown functionality

**Width Variants:**
- `MIN-WIDTH: 280PX` - Minimum column width
- `MIN-WIDTH: 220PX` - Compact minimum width  
- `MIN-WIDTH: 172PX` - Extra compact width
- `FILL` - Flexible width that fills available space
- `MAX-WIDTH: 120PX` - Maximum constrained width
- `MAX-WIDTH: 40PX` - Icon/action column width

##### **2. Header Component Variants**

![Image](https://github.com/user-attachments/assets/5f9b9b49-ee06-4b28-bd35-263991c17703)

- `DEFAULT` (40px height) - Standard header
- `SMALL` (32px height) - Compact header

##### **3. Cell Component Variations**

![Image](https://github.com/user-attachments/assets/e6a2db21-965c-4008-8c4a-0fdddd8d4344)

Support for different cell content types:
- **User/DAO**: Avatar + name display
- **Title**: Text with optional info icon
- **Badge**: Status indicators
- **Text**: Simple text content
- **Variation**: Percentage with trend indicators
- **Chart**: Inline chart visualization
[...]

PR #1345, opened 2025-09-18
feat: design-system table

brunod-e:
Related issue https://github.com/blockful/anticapture/issues/848

Things to consider:
- Filter button will be made in another task
- Width variants didn't make sense, cell must not have fixed width to make it responsible
- As discussed with Isa, Cell components will be made in another task like ui-commons

Features included in this task:
- Infinite scroll
- Size variants
- Download as csv


edulennert (changes requested):
Loading State shows variation as 0:
image

Width of the table changes during loading
On Delegates table, activity is shown as 0/0 after loading, and after some time the real activity is shown:
https://www.loom.com/share/d604a59bfc9743f8aa0211ac358d0b79?sid=ee5c3eb3-91c6-47f3-a0f0-77a0699e8953

It would be nice to have consistent loading, no results and error states.

Infinite Scroll doesn't work on mobile

edulennert (changes requested):
When the Table is loading we still have a "no items found" or errors across tables, the loading state of the table should be shown.

image

Also, there is a couple of layout shifts when the Delegate votingPower table is loading.

This here is not 100% important, but this micro space at the top of the table header is bugging me lol
image

isadorable-png (approved):

PR #1357, opened 2025-09-24
Feat/1242 feat proposal votes

edulennert:


isadorable-png (commented):
- Main table should not have border
image


- Address column should not have dashed border-bottom
- there's no border bottom or top on the table while scrolling, also the label bar is not fixed (sticky) as the user scrolls
- I think we can distribute more equally each column width
- set all the rows text to font-weight regular
- we can also set bigger fixed height for this table, maybe something around 800px

image

PR #1368, opened 2025-10-03
Proposal non voters

pikonha:


apps/indexer/src/api/repositories/drizzle.ts
```diff
@@ -156,6 +175,136 @@ export class DrizzleRepository {
...
+    limit: number,
+    orderDirection: "asc" | "desc",
+    addresses?: Address[],
+  ): Promise<{ voter: Address; votingPower: bigint }[]> {
+    return await db
+      .select({
+        voter: accountPower.accountId,
+        votingPower: accountPower.votingPower,
+      })
+      .from(accountPower)
```
LeonardoVieira1630, line 190: This account power is time sensitive? I mean, the current account power is being used right? But a "possible voter" needs to be some one that has voting power in the "timestamp" of the proposal. Isnt it? Like, if i had 0_vp in the beginning of the prop, but now i have  1k_vp, im not a "possible voter" right?

apps/indexer/src/api/services/proposals/proposals.ts
```diff
@@ -104,4 +119,47 @@ export class ProposalsService {
...
+      proposalId,
+      skip,
+      limit,
+      orderDirection,
+      addresses,
+    );
+
+    const _addresses = addresses ? addresses : nonVoters.map((v) => v.voter);
+
+    const comparisonTimestamp = Math.floor(Date.now() / 1000 - DaysEnum["30d"]);
```
LeonardoVieira1630, line 143: "DaysEnum["30d"]" smells strange uheuehuheueh. Is it correct? why do we have it?

apps/indexer/src/api/repositories/drizzle.ts
```diff
@@ -1,7 +1,26 @@
...
+  desc,
+  eq,
+  gte,
+  gt,
+  inArray,
+  sql,
+  isNull,
+  count,
+  max,
+} from "ponder";
```
LeonardoVieira1630, line 14:
```suggestion
import * as operators from "ponder";
```

apps/indexer/src/api/repositories/drizzle.ts
```diff
@@ -156,6 +175,136 @@ export class DrizzleRepository {
...
+  ): Promise<Record<Address, bigint>> {
+    const timestamps = await db
+      .select({
+        voterAccountId: votesOnchain.voterAccountId,
+        lastVoteTimestamp: max(votesOnchain.timestamp),
+      })
+      .from(votesOnchain)
+      .where(inArray(votesOnchain.voterAccountId, voters))
+      .groupBy(votesOnchain.voterAccountId)
+      .orderBy(desc(max(votesOnchain.timestamp)));
```
LeonardoVieira1630, line 242:
```suggestion
```

i think we dont need this operation. Like, this func is used on getProposalNonVoters. Over there we already order.

apps/indexer/src/api/services/proposals/proposals.ts
pikonha: 30 days in seconds, not beautiful but it works

apps/indexer/ponder.schema.ts
```diff
@@ -157,14 +157,14 @@ export const transfer = onchainTable(
...
     voterAccountId: drizzle.text("voter_account_id").$type<Address>().notNull(),
-    proposalId: drizzle.text("proposal_id"),
-    support: drizzle.text(),
-    votingPower: drizzle.text(),
-    reason: drizzle.text(),
-    timestamp: drizzle.bigint(),
+    proposalId: drizzle.text("proposal_id").notNull(),
+    support: drizzle.text().notNull(),
+    votingPower: drizzle.bigint().notNull(),
+    reason: drizzle.text().notNull(),
```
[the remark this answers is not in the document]
pikonha: nice catch, I'll remove it

apps/indexer/src/api/services/proposals/proposals.ts
PedroBinotto, line 143: The following snippet:

```ts
Math.floor(Date.now() / 1000)
```

Is repeated quite a bit around the codebase; a couple of questions:

1. Why is this necessary? Does it have to do with the way `drizze` serializes/deserializes timestamps?
2. Wouldn't it be better to extract this logic to a helper function in an `utils` file somewhere?

apps/indexer/src/api/repositories/drizzle.ts
```diff
@@ -156,6 +175,136 @@ export class DrizzleRepository {
     return db.$count(proposalsOnchain);
   }
 
+  async getProposalNonVoters(
+    proposalId: string,
+    skip: number,
+    limit: number,
+    orderDirection: "asc" | "desc",
```
PedroBinotto, line 182: Maybe we could put these literals in a constant/enum at some point, as they are re-declared a lot around the project

PedroBinotto (commented):
lgtm

PR #1367, opened 2025-10-03
[DEV-61] Feat/add token value to endpoint

PedroBinotto:
### changelog

- Added `getTokenProperties` operation:

# `GET /token/properties`

**Summary:** Get token properties  
**Description:** Get property data for a specific token  
**Tags:** `tokens`

---

## Query Parameters

| Name     | In    | Type   | Required | Allowed Values | Default |
|----------|-------|--------|----------|----------------|---------|
| currency | query | string | No       | `eth`, `usd`   | `usd`   |

---

### **200 OK**

Returns data concerning token-specific values and attributes; ex.:

```json
{
  "id": "string",
  "name": "string | null",
  "decimals": 18,
  "totalSupply": "string",
  "delegatedSupply": "string",
  "cexSupply": "string",
  "dexSupply": "string",
  "lendingSupply": "string",
  "circulatingSupply": "string",
  "treasury": "string",
  "price": 123.45
}
```

<details>
<summary>Show schema</summary>

```yaml
  /token/properties:
    get:
      operationId: tokenProperties
      summary: Get token properties
      description: Get property data for a specific token
      tags:
        - tokens
      parameters:
        - schema:
            type: string
            enum:
              - eth
              - usd
            default: usd
          required: false
          name: currency
          in: query
      responses:
        '200':
          description: Returns data concerning token-specific values and attributes
          content:
            application/json:
              schema:
                type: object
                properties:
                  id:
                    type: string
                  name:
                    type: string
                    nullable: true
                  decimals:
                    type: number
                  totalSupply:
                    type: string
                  delegatedSupply:
                    type: string
                  cexSupply:
                    type: string
                  dexSupply:
                    type: string
                  lendingSupply:
                    type: string
                  circulatingSupply:
                    type: string
                  treasury:
                    type: string
                  price:
                    type: number
                required:
                  - id
                  - name
                  - decimals
                  - totalSupply
                  - delegatedSupply
                  - cexSupply
                  - dexSupply
                  - lendingSupply
                  - circulatingSupply
                  - treasury
                  - price
        '404':
          description: Token not found
```
</details>

PR #1372, opened 2025-10-06
Remove time switcher from panel

PedroBinotto:


pikonha (changes requested):
Wait for [this other PR](https://github.com/blockful/anticapture/pull/1367) to change the way the delegation supply is being gathered

PR #1393, opened 2025-10-16
[DEV-106] Refactor transaction query to enable pagination

PedroBinotto:
# Changelog

### Added
- **Date range filtering**: Added `fromDate` and `toDate` query parameters to the transactions endpoint for filtering by timestamp ranges
- **Validation**: Added validation to ensure `fromDate` is less than or equal to `toDate`
- **Performance optimization**: Implemented optimized SQL queries for filtered vs unfiltered transactions
  - Filtered queries use CTE-based aggregation
  - Unfiltered queries use lateral joins for better performance

### Changed
- **Amount formatting**: Transaction and delegation amounts now returned in ETH (using `formatEther`) instead of wei
- **Repository architecture**: Consolidated transfer and delegation repositories into a single `TransactionsRepository` with unified SQL queries
- **Query strategy**: Automatic query selection based on whether filters are applied using new `containsAnyValue` utility

### Removed
- **Redundant repositories**: Removed separate `TransfersRepository` and `DelegationsRepository` in favor of unified approach

### Fixed
- **Transactions query**: The new query implementations allow for native pagination and provide better performance due to the separate routines for filtered and unfiltered querying and pure SQL implementation.

### DIFF
--- a/apps/indexer/src/api/services/transactions/index.ts
+++ b/apps/indexer/src/api/services/transactions/index.ts
@@ -4,7 +4,7 @@ import {
   TransactionsResponse,
   TransactionMapper,
 } from "../../mappers/transactions";
-import { containsAnyValue } from "@/lib/utils";
+import { containsAnyValue } from "../utils";
 
 export class TransactionsService {
   constructor(private transactionsRepository: TransactionsRepository) {}
--- a/apps/indexer/src/api/services/utils.ts
+++ b/apps/indexer/src/api/services/utils.ts
@@ -1,0 +1,9 @@
+export function containsAnyValue(data: Record<string, unknown>): boolean {
+  return Object.keys(data).some((key) => {
+    const value = data[key];
+    if (value === null || value === undefined) return false;
+    if (typeof value === "object" && Object.keys(value).length === 0)
+      return false;
+    return true;
+  });
+}
--- a/apps/indexer/src/lib/utils.ts
+++ b/apps/indexer/src/lib/utils.ts
@@ -30,13 +30,3 @@ export function max(...values: bigint[]): bigint {
 export function getChain(chainId: number): chains.Chain | undefined {
   return Object.values(chains).find((chain) => chain.id === chainId);
 }
-
-export function containsAnyValue(data: Record<string, unknown>): boolean {
-  return Object.keys(data).some((key) => {
-    const value = data[key];
-    if (value === null || value === undefined) return false;
-    if (typeof value === "object" && Object.keys(value).length === 0)
-      return false;
-    return true;
-  });
-}


apps/indexer/src/api/repositories/transactions.repository.ts
```diff
@@ -1,24 +1,189 @@
...
+          'isCex', ft.is_cex,
+          'isDex', ft.is_dex,
+          'isLending', ft.is_lending,
+          'isTotal', ft.is_total
+        )
+      )
+      FROM filtered_transfers ft
+      WHERE ft.transaction_hash = ${transaction.transactionHash}),
+      ARRAY[]::jsonb[]
+    )`.as("transfers"),
```
PedroBinotto: I really wish we could simplify this method like this:

```diff
-      (SELECT ARRAY_AGG(
-        jsonb_build_object(
-          'transactionHash', ft.transaction_hash,
-          'daoId', ft.dao_id,
-          'tokenId', ft.token_id,
-          'amount', ft.amount,
-          'fromAccountId', ft.from_account_id,
-          'toAccountId', ft.to_account_id,
-          'timestamp', ft.timestamp,
-          'logIndex', ft.log_index,
-          'isCex', ft.is_cex,
-          'isDex', ft.is_dex,
-          'isLending', ft.is_lending,
-          'isTotal', ft.is_total
-        )
-      )
-      FROM filtered_transfers ft
-      WHERE ft.transaction_hash = ${transaction.transactionHash}),
+      (SELECT ARRAY_AGG(to_jsonb(ft.*))
+       FROM filtered_transfers ft
+       WHERE ft.transaction_hash = ${transaction.transactionHash}),
```

And it would be possible were it not for the difference between the column name case (`snake_case`) and the object attributes names (`camelCase`); as that's the case, I think it's more technically correct to keep the return type of the method "pure" (i.e.: to have the method return the table datatype) than to have some sort of conversion here to save a couple of lines of code

apps/indexer/src/api/repositories/transactions.repository.ts
PedroBinotto: https://explain.dalibo.com/plan/3634ff2c79cc214b

apps/indexer/src/api/mappers/transactions.ts
pikonha:
```suggestion
    .transform((affectedSupply) => affectedSuplly.map(s => s.toUpperCase())
    .transform((affectedSupply) => {
```

pikonha (commented):
we need to handle `fromDate` and `toDate` params to enable custom chart integration

apps/indexer/src/api/repositories/transactions.repository.ts
```diff
@@ -1,24 +1,79 @@
...
-      with: {
-        transfers: true,
-        delegations: true,
-      },
-    });
+  async getFilteredAggregateTransactions(
+    filter: TransactionsRequest,
+  ): Promise<DBTransaction[]> {
+    const { transfer: transferFilter, delegation: delegationFilter } =
+      toSql(filter);
```
pikonha: could you move this implementation to this repo file

pikonha (approved):

apps/indexer/src/api/repositories/transactions.repository.ts
```diff
@@ -1,24 +1,241 @@
...
-        delegations: true,
-      },
-    });
+  async getFilteredAggregateTransactions(
+    filter: TransactionsRequest,
+  ): Promise<DBTransaction[]> {
+    const { transfer: transferFilter, delegation: delegationFilter } =
+      this.filterToSql(filter);
+
+    const query = sql`
```
PedroBinotto: https://explain.dalibo.com/plan/9gae943ac2fda435

PR #1698, opened 2026-03-10
Feat/structured logging

PedroBinotto:
## Changelog

- Enable structured logging with PinoJS + log tracking, querying and aggregation through OpenTelemetry -> Loki -> Grafana;
- Distinguish between client (default) and notification-system requests via custom header;
- Add per-layer tracing of backend services (Controller-Service-Repository span wrapping);


Dockerfile.indexer
```diff
@@ -17,6 +17,7 @@ COPY --from=builder /app/turbo.json turbo.json
 RUN pnpm --filter @anticapture/observability build
 
 FROM base AS runner
+ENV NODE_ENV=production
```
pikonha, line 20: this is gonna be use on dev as well

Dockerfile.api
```diff
@@ -18,6 +18,7 @@ RUN pnpm --filter @anticapture/observability build
 RUN pnpm run api build
 
 FROM base AS runner
+ENV NODE_ENV=production
```
pikonha, line 21: this is gonna be use on dev as well

Dockerfile.indexer
PedroBinotto: the `dev` deploy env overrides this and sets `NODE_ENV=dev`; from what I've checked, setting the default to `production` in the Dockerfile and overriding it in dev is the industry standard, at least for NodeJS backends.

pikonha (approved):

PR #1700, opened 2026-03-11
Gateful

LeonardoVieira1630:
## REST Gateway

Adds a new `rest-gateway` package — a Hono-based API gateway that aggregates multiple upstream DAO APIs into a unified REST endpoint.

### What's included

- **Proxy route**: forwards requests to upstream APIs
- **DAOs route**: aggregates DAO data from all upstreams
- **Delegation route**: aggregates and merges delegation data
- **Address enrichment route**: enriches address data across upstreams
- **Health check route**
- **OpenAPI docs merging** from upstream APIs
- **Shared utilities**: config, fan-out helpers
- Tests for all routes and services


pikonha (commented):
## Code Review

Good architecture — clean separation of concerns with routes, services, and shared utilities. The proxy pattern and fan-out approach are solid. A few issues to address before merge.

apps/rest-gateway/src/index.ts
```diff
@@ -0,0 +1,54 @@
...
+addressEnrichment(app, config.addressEnrichmentUrl);
+
+// OpenAPI docs
+app.get("/docs/json", async (c) => {
+  const ownSpec = app.getOpenAPI31Document({
+    openapi: "3.1.0",
+    info: { title: "Anticapture REST Gateway", version: "1.0.0" },
+  });
+
+  const merged = await mergeUpstreamDocs(ownSpec, config.daoApis);
```
pikonha: **No caching on OpenAPI doc merge.** `mergeUpstreamDocs` fetches from all upstreams on every request to `/docs/json`. Consider caching the result with a short TTL (e.g. 60s).

alextnetto (commented):
## Code Review — PR #1700 "Gateful" (REST Gateway)

Good architectural direction — moving from GraphQL Mesh to a lean Hono REST gateway is a solid simplification. Clean separation of concerns with route/service split, good use of `Promise.allSettled` for graceful degradation, and OpenAPI-first design with `@hono/zod-openapi`. Several issues need attention before merge.

---

### Critical (Must Fix)

**1. Wrong upstream endpoint path in `DelegationService`**
`src/delegation/service.ts:35` — calls `/delegation-percentage-by-day` but the upstream API exposes `/delegation-percentage` (see `apps/api/src/controllers/delegation-percentage/index.ts`). Every delegation fan-out will 404, `Promise.allSettled` silently drops failures, and the gateway returns a misleading empty response. This was flagged by Codex and appears still unresolved.

**2. Index-based aggregation can produce `NaN` on misaligned arrays**
`src/delegation/service.ts:~101` — `aggregateMeanPercentage` iterates by array index. If DAOs return different numbers of items (e.g., one DAO has a data gap), `response.items[index]` is `undefined` → `parseFloat(undefined)` → `NaN` silently corrupts the mean. The "guaranteed consecutive daily data" assumption from the old api-gateway is undocumented and fragile. Either join by date key (map-based), or add a runtime assertion that all arrays have equal length after alignment.

**3. Missing `typecheck` and `lint` scripts in `package.json`**
Per CLAUDE.md, every package must support `pnpm <package> typecheck` and `pnpm <package> lint`. Without these, the code can't be verified against the project's quality gates. Add `typecheck`, `lint`, and `lint:fix` scripts plus eslint/prettier devDependencies.

---

### Important (Should Fix)

**4. `noImplicitAny: false` in `tsconfig.json:14`**
This contradicts `strict: true` and weakens type safety. The old api-gateway needed this for Mesh-generated code, but this package is written from scratch — no reason to carry the weakness forward. Remove it and fix any resulting type errors.

**5. Zod v4 vs monorepo v3 mismatch** (`package.json` — `zod: ^4.3.6`)
Every other package uses Zod v3. This fragments the dependency graph and `@hono/zod-openapi` jumped from `^0.19.6` to `^1.2.2` to match. Either downgrade to v3 or document the migration plan.

**6. `limit=0` still treated as falsy** (`delegation/service.ts:37`)
`if (args.limit) params.set(...)` drops `limit=0` since `0` is falsy. The route's regex accepts `"0"`, so this is reachable. Use `if (args.limit !== undefined)` or reject `0` at the route validation layer.

**7. CORS wildcard `origin: "*"` in production** (`index.ts:20`)
Acceptable for dev, but should be configurable via env var for production.

**8. No timeout on OpenAPI doc fetches** (`upstream-docs.ts`)
`fanOutGet` correctly uses `AbortSignal.timeout(5000)`, but `fetchDoc` in `upstream-docs.ts` has no timeout. A hung upstream blocks `/docs/json` indefinitely. Add `signal: AbortSignal.timeout(5000)`.

**9. Proxy body forwarding not verified** (`proxy/route.ts`)
The test mocks fetch, so POST body forwarding is never actually tested. Verify with `body: c.req.raw.body` explicitly, or add an integration test.

---

### Suggestions

- `openapi-types` in devDependencies is unused (code uses `openapi3-ts/oas31`)
- Health endpoint returns `null` body — consider `{ status: "ok" }` for monitoring tools
- No structured logging — given the otel-instrumentation branch exists, consider using a structured logger from the start
- Tests don't assert the actual upstream URL called, which is why bug #1 slips through

---

### Prior Review Status

| Issue | Status |
|-------|--------|
| ✅ BigInt crash (pikonha) | Fixed — regex validates before BigInt |
| ✅ No URL validation on DAO APIs (pikonha) | Fixed — `z.url()` + `safeParse` |
| ✅ Fetch timeouts (pikonha) | Fixed in `fanOutGet`, missing in `fetchDoc` |
| ✅ Invalid delegation params (codex) | Fixed — regex validation added |
| ⚠️ `limit=0` falsy (pikonha) | Partially fixed — `buildPaginatedResponse` uses `??`, but `params.set` still uses truthiness |
| ❌ Wrong endpoint path (codex) | Unresolved |
| ❌ Index-based aggregation (pikonha/codex) | Unresolved — alignment added but still fragile |
| ❌ `noImplicitAny: false` (pikonha) | Unresolved |
| ❌ Zod v4 mismatch (pikonha) | Unresolved |
| 📋 OpenAPI doc caching (pikonha) | Acknowledged as next task |

**Recommendation:** Fix the critical items (#1 wrong endpoint path, #2 aggregation safety, #3 missing scripts) before merge.

LeonardoVieira1630:
@alextnetto , all that were already solved. Your branch were not up to date.

pikonha (approved):

Issue #1, opened 2024-08-12
Create and config Project's board


LeonardoVieira1630:
Set up a comprehensive task tracking board. This includes creating the board structure, defining individual tasks, organizing these tasks into sprints, and establishing key milestones to monitor progress. 
- [x]  Create board
- [ ]  Create tasks
- [x]  Create sprints
- [x]  Create milestones

PR #1701, opened 2026-03-11
[DEV-475] Composite dao endpoint for header metrics

PedroBinotto:


alextnetto (commented):
## Code Review — PR #1701 "[DEV-475] Composite DAO Endpoint"

Good refactoring — extracting `GovernanceActivityService` from the controller, introducing the `fetchGovernanceData` flag, and moving from `DaysEnum` to `fromDate` timestamps are all clean improvements. Two critical bugs and one runtime validation issue need attention.

---

### Critical (Must Fix)

**1. Malformed SQL in `getAverageTurnoutCompare`** (`repositories/drizzle/index.ts`)
The `current_average_turnout` CTE contains:
```sql
WHERE ${proposalsOnchain.timestamp} >= ${proposalsOnchain.timestamp} >= ${fromDate}
```
This is a chained comparison: `("timestamp" >= "timestamp")` → `true`, then `true >= fromDate` → boolean-to-integer comparison. PostgreSQL will either error or return wrong results. Every code path through `getAverageTurnoutCompare` (governance activity, `/dao?fetchGovernanceData=true`) is affected.

**Fix:** Should be:
```sql
WHERE ${proposalsOnchain.timestamp} >= ${fromDate}
```

**2. Gateway resolver forwards parent `info` to delegated `dao` calls** (`api-gateway/src/resolvers/daos.ts`)
The `info` object belongs to the `daos` field (`DAOList`), but it's forwarded to `client.dao(...)` which resolves `dao_200_response`. In Mesh delegation, this causes the wrong selection set to be projected. The previous code intentionally omitted `info`. Unless specifically required by a Mesh version change, remove it.

---

### Important (Should Fix)

**3. `fetchGovernanceData` enum case mismatch** (`meshrc.ts` + `api/src/controllers/dao/index.ts`)
The GraphQL enum uses `TRUE`/`FALSE` (uppercase), but the API's Zod schema expects `"true"`/`"false"` (lowercase):
```typescript
fetchGovernanceData: z.enum(["true", "false"]).default("false")
```
Mesh forwards the enum value as-is → Zod rejects `"TRUE"` → 400 error. Either accept both cases in Zod (`.transform(v => v.toLowerCase())`), change the enum values to lowercase, or add case-insensitive validation.

**4. Zero-turnout truthiness**
`normalizedTurnout` uses truthiness to check `currentAverageTurnout`. Since this comes as a string (`"0"` is truthy), it won't cause issues for the current data flow. However, `calculateQuorumGap` receives `avgTurnout = 0` and computes `-100` (0/quorum - 1). Confirm this is the intended behavior or add an explicit `avgTurnout === 0 → NOT_APPLICABLE` guard.

**5. `quorumGap` truthiness in mapper** (`mappers/dao/index.ts`)
If `dbData.quorumGap` is exactly `0` (turnout equals quorum), the truthiness check converts it to `NOT_APPLICABLE` instead of `"0"`. Use `dbData.quorumGap !== null && dbData.quorumGap !== undefined` instead of a truthiness check.

**6. Duplicate `getVotesCompare` in `GovernanceActivityRepository` interface** (`services/governance-activity/index.ts`)
The method is declared twice. TypeScript allows it but it's confusing. Remove the duplicate.

---

### Suggestions

- Cache is correctly scoped to RPC parameters only (not governance data), but a doc comment on `resolveRPCData` clarifying this would prevent future confusion
- `DummyGovernanceActivityService` in `cmd/aave.ts` is a pragmatic solution — consider extracting to a shared null-object if more DAOs need it
- `GovernanceActivityService` constructor params should be `private readonly` for consistency
- Schema descriptions were degraded from `"""triple-quoted"""` to single-line strings with embedded `\n` and localhost URLs — this appears to be a schema regeneration artifact worth cleaning up

**Recommendation:** Fix #1 (SQL bug) and #3 (case mismatch) before merge — both cause runtime failures.

apps/api/src/mappers/dao/index.ts
```diff
@@ -1,6 +1,19 @@
...
+
+export const DaoParametersRequestQuerySchema = z.object({
+  fromDate: z
+    .string()
+    .transform((val) => Number(val))
+    .optional(),
+  fetchGovernanceData: z
+    .enum(["true", "false"])
+    .default("false")
+    .transform((val) => val === "true"),
```
pikonha:
```suggestion
  fetchGovernanceData: z
    .enum(["true", "false"])
    .default("false")
    .transform((val) => val === "true")
    .openapi({type: "boolean"}),
```

PR #1705, opened 2026-03-12
Feat/auth & ipv6

LeonardoVieira1630:
- [ x ] Add auth token
- [ x ] Add dns ipv6 support


alextnetto (commented):
## Code Review — PR #1705 "Auth & IPv6"

The auth and DNS changes are well-motivated for Railway deployment. The `alreadySupportCalldataReview` field is cleanly threaded through the full stack. Two critical deployment issues need fixing.

---

### Critical (Must Fix)

**1. Health check not exempt from bearer auth** (`rest-gateway/src/index.ts:28-30`)
`app.use("*", bearerAuth(...))` is registered before `health(app)`. Hono evaluates middleware in registration order, so `/health` requires a token when `BLOCKFUL_API_TOKEN` is set. This breaks load balancer health probes, Railway liveness checks, and uptime monitors.

**Fix:** Either register the health route *before* the auth middleware, or scope auth to exclude `/health`:
```typescript
health(app); // Register health first
if (config.blockfulApiToken) {
  app.use("*", bearerAuth({ token: config.blockfulApiToken }));
}
```

**2. IPv6-only bind may fail on IPv4-only hosts** (`rest-gateway/src/index.ts:63`)
`hostname: "::"` works on dual-stack systems but fails with `EAFNOSUPPORT` where IPv6 is disabled (some CI runners, Docker containers with IPv6 disabled). Make it configurable:
```typescript
// In config.ts
HOST: z.string().default("::"),
// In index.ts
serve({ fetch: app.fetch, port: config.port, hostname: config.host });
```

---

### Important

**3. Offchain query types added without clear resolution path**
`schema.graphql` adds `offchainProposals`, `offchainProposalById`, `votesOffchain`, `votesOffchainByProposalId` plus ~80 lines of types. These may auto-resolve via GraphQL Mesh, but there's no corresponding rest-gateway route. If these need fan-out aggregation like `/daos`, that work is missing. Clarify the intent.

**4. PR scope is broad — four unrelated concerns**
Bearer auth + IPv6 (rest-gateway infra), `alreadySupportCalldataReview` (API layer), offchain queries (api-gateway schema). These are independent features that would be easier to review/revert/bisect as separate PRs.

**5. Single static bearer token is minimal auth**
Fine as a first iteration for internal traffic, but document it as such. All clients share one token, no rotation without redeploy, no per-client identification.

---

### Suggestions

- `alreadySupportCalldataReview` naming — consider `supportsCalldataReview` or `calldataReviewEnabled` for clarity
- The fan-out `console.error` addition in `fan-out.ts` is a good operational improvement
- The DNS `verbatim` comment explaining Railway AAAA records is excellent documentation

**Recommendation:** Fix critical items (#1 health exemption, #2 IPv6 fallback) before merge.

pikonha (approved):

PR #1733, opened 2026-03-18
feat: enhance grafana

PedroBinotto:
This fixes distributed tracing across services (waterfalling);

**Before:** tracerProvider.register() — relied on SDK defaults.

**After:** registers with an explicit CompositePropagator containing W3CTraceContextPropagator, and adds @opentelemetry/core as a direct dependency to import those classes.

This ensures traceparent/tracestate HTTP headers are propagated between services, which is required for distributed trace correlation in Grafana Tempo. The rest of the diff is lockfile nois from a pnpm install.


PedroBinotto:
@claude

PedroBinotto:
@claude

pikonha (approved):

PR #1756, opened 2026-03-20
fix: load root .env for all services and fix OTel ESM/CJS compat

alextnetto:
## Summary
- Add `dotenv --` to all pnpm scripts in root `package.json` so the root `.env` is always injected into turbo child processes when running from the repo root
- Bundle OpenTelemetry dependencies into `@anticapture/observability` via a `tsup.config.ts` to fix the ESM/CJS resolution conflict in `@opentelemetry/resources@1.30.x` that broke ponder and tsx consumers
- Keep `@opentelemetry/api` external to preserve singleton behavior
- Add missing `apps/gateful/.env.example` documenting required env vars

## Environment loading convention

Each app calls `dotenv.config()` in its entry point, which loads `.env` relative to the current working directory (`process.cwd()`). Combined with `dotenv-cli` in the root scripts, this gives us:

| How you run | Root `.env` loaded by | App `.env` loaded by | Result |
|---|---|---|---|
| `pnpm api dev` (from root) | `dotenv-cli` | `dotenv.config()` (no-op, won't overwrite) | Root `.env` used |
| `cd apps/api && pnpm dev` (from app) | Not invoked | `dotenv.config()` loads `apps/api/.env` | App `.env` used |

This means developers can override specific vars per-app by creating a `.env` in the app folder (copy from `.env.example`), while the root `.env` remains the single source of truth for running from the repo root.

## Test plan
- [x] `pnpm api dev -- ENS` starts successfully
- [x] `pnpm gateway dev` starts and connects to API
- [x] `pnpm dashboard dev` starts on Next.js
- [x] `pnpm indexer dev` starts ponder, connects to DB and RPC
- [x] Typecheck and lint pass on observability package
- [ ] `pnpm client codegen` loads root `.env` (new `dotenv --`)
- [ ] `pnpm offchain-indexer dev` loads root `.env` (new `dotenv --`)
- [ ] `pnpm address dev` loads root `.env` (new `dotenv --`)


packages/observability/package.json
PedroBinotto, line 1: the changes to the `observability` don't look right; these changes are exclusively for the dev setup workflow (local), right? can you please paste the errors you were getting that required this?

PedroBinotto (changes requested):

PR #1768, opened 2026-03-20
feat: Lil Nouns DAO support + dev.sh rewrite

alextnetto:
## Summary

- **Lil Nouns DAO integration**: indexer (token + governor handlers), API (contracts, client routing, NFT pricing, event relevance), and dashboard config
- **Configurable initial page per DAO**: each DAO can define which page loads first (overview vs governance)
- **dev.sh rewrite**: single `pnpm dev` command runs the full local stack (indexer, API, gateway, gateful, client, dashboard) with root `.env` as the source of truth
  - All ports configurable via `PORT_*` env vars
  - Kills stale processes on startup
  - DB `search_path` opt-in via `DB_SCHEMA` env var instead of hardcoded

## Test plan

- [ ] Run `pnpm dev` — all 6 services start with correct DAO from root `.env`
- [ ] Query gateway: `curl -H "anticapture-dao-id: LIL_NOUNS" http://localhost:4000/graphql?query={proposals{items{id title}}}`
- [ ] Dashboard loads and hits gateway at `localhost:4000`
- [ ] Lil Nouns proposals, governance, and token data render correctly


apps/api/src/env.ts
```diff
@@ -7,9 +7,10 @@ dotenv.config();
 
 const envSchema = z.object({
   RPC_URL: z.string(),
-  DATABASE_URL: z
-    .string()
-    .transform((val) => `${val}?options=-c%20search_path%3Danticapture`),
+  DATABASE_URL: z.string().transform((val) => {
+    const schema = process.env.DB_SCHEMA ?? "anticapture";
```
alextnetto: not needed, can revert

apps/api-gateway/schema.graphql
alextnetto, line 1: You changed to much here, what happened? you should change as less as possible, might be removing things that are needed

apps/indexer/config/lil-nouns.config.ts
alextnetto, line 1: number of file should be lil-nouns or lilNouns

apps/indexer/src/indexer/lilnouns/index.ts
```diff
@@ -0,0 +1,4 @@
+export * from "./abi";
+
```
alextnetto: delete this whitespace

apps/indexer/package.json
```diff
@@ -4,7 +4,7 @@
   "private": true,
   "scripts": {
     "dev": "ponder dev",
-    "start": "ponder start --views-schema=anticapture --schema=$RAILWAY_DEPLOYMENT_ID",
+    "start": "ponder start --views-schema=anticapture --schema=vasco",
```
alextnetto: put a proper schema name for running it locally

PR #1789, opened 2026-03-27
Feat: integrate Gnosis dashboard

PedroBinotto:


isadorable-png:
## 🔍 UI Review — `Feat: integrate Gnosis dashboard` [DRAFT]

> **Figma**: Path A — no Figma or ClickUp links found. Reviewed against codebase patterns and UX heuristics.
> **Preview**: `anticapture-git-feat-index-gnosis-v3-ful.vercel.app` — auth-gated (403).

---

### 🟡 Findings

#### 1. `<a><Button>` nested interactive elements `[UX-heuristic]`

**Severity: Medium** — invalid HTML; some screen readers and browsers disagree on which element fires the click.

In `ProposalSection.tsx` (mobile bottom bar for offchain) and `ProposalHeader.tsx` (desktop "Vote on Snapshot"):

```tsx
<a href={snapshotLink} target="_blank" rel="noopener noreferrer" className="flex w-full">
  <Button className="flex w-full">
    <ExternalLink className="size-3.5" />
    Vote on Snapshot
  </Button>
</a>
```

`<button>` inside `<a>` is invalid HTML per the spec. Use the DS `Button`'s `asChild` prop instead:

```tsx
<Button asChild variant="outline" size="sm" className="flex w-full">
  <a href={snapshotLink} target="_blank" rel="noopener noreferrer">
    <ExternalLink className="size-3.5" />
    Vote on Snapshot
  </a>
</Button>
```

---

#### 2. `Number(supportValue)` coerces `null` to `0` `[Code-only]`

**Severity: Medium** — `0` is typically "For" in on-chain vote support values. This means a user who hasn't voted will appear to have voted "For."

In `ProposalHeader.tsx`:

```tsx
// Before
supportValue={supportValue ?? undefined}
// After
supportValue={Number(supportValue)}
```

`Number(null)` returns `0`. If `ProposalHeaderAction` uses `supportValue` to determine vote display state (and index 0 = "For"), passing `0` instead of `undefined`/`null` could render the "Cast your vote" UI incorrectly for wallets with no vote on record. Worth verifying that `ProposalHeaderAction` handles `0` vs `undefined` distinctly.

---

#### 3. `<h1>` used as dropdown item label `[Code-only]`

**Severity: Low** — semantically wrong heading element inside a `role="menu"`.

In `HeaderDAOSidebarDropdown.tsx`:

```tsx
<h1 className={cn("text-primary text-sm font-normal")}>
  {item.label}
</h1>
```

`<h1>` is a document-level heading. Inside a dropdown menu item it creates an extra heading landmark accessible users won't expect. Use `<span>` or `<p>`.

---

#### 4. `GnosisIcon` background vs `gno.ts` `svgBgColor` mismatch `[Code-only]`

**Severity: Low** — potential visual inconsistency in avatar/badge contexts that use the config color.

`gno.ts` declares `svgBgColor: "#fff"` (white), but `GnosisIcon.tsx` renders `<rect ... fill="#00193C" />` (dark navy). If any component uses `svgBgColor` to compute the icon's container or fallback background, Gnosis's dark icon would appear on a white background rather than the navy it was designed for. Compare: `torn.ts` sets `svgBgColor: "#1a1a2e"`, which matches the `TornadoCashIcon` rect.

---

### ✅ Confirmed correct

- **`proposalTypeConfiguration` migration** — The three-way `"onchain" | "offchain" | "composite"` config cleanly replaces `offchainProposals: boolean`. ENS is correctly marked `"composite"`, GNO `"offchain"`. Tab group only shown for `"composite"`. Default tab resolves to `"offchain"` for GNO (via `withDefault(hasOnchain ? "onchain" : "offchain")`). ✓
- **`skip: !hasOnchain` in `useProposals`** — Prevents unnecessary GraphQL query for offchain-only DAOs. ✓
- **`choices[choice]` fix in `OffchainVotesContent`** — Corrects 1-indexed → 0-indexed Snapshot array access. ✓
- **`OffchainVotingModal` removed** — Redirecting to Snapshot rather than hosting a replica is cleaner and reduces the surface area for voting bugs. ✓
- **`GnosisIcon.tsx`** — Follows `DaoIconProps` pattern, `showBackground` prop correctly wraps `<rect>`. ✓
- **Storybook category renames** (`Data Display/Drawer` → `Feedback/Drawer`) — Consistent with other overlay components. ✓
- **`fade-out` animation removed from `globals.css`** — The only consumer (`HeaderDAOSidebarDropdown`'s `isClosing` state) was removed in the same PR. Clean. ✓
- **`Select`/`RadioCard` removed from `form/fields/index.ts`** — If the barrel re-export is gone but consumers still work, these were either moved to direct imports or removed. Consistent with the rest of this refactor. ✓
- **`HeaderDAOSidebarDropdown` ARIA** — `role="menuitemradio"` + `aria-checked` is more accurate than plain `role="menuitem"` for a DAO selector. ✓

---

*Reviewed by @isadorable-png — automated UI review*

---

PR #1794, opened 2026-03-31
Feat/offchain non voters

LeonardoVieira1630:
## Summary

Add a new API endpoint (`GET /offchain/proposals/{id}/non-voters`) that returns active delegates who did not vote on a given offchain (Snapshot) proposal, enabling governance participation analysis. This also normalizes addresses to checksummed format in the offchain indexer to ensure consistent cross-schema joins.

## Changes

- **API endpoint & controller** (`apps/api/src/controllers/votes/offchainNonVoters.ts`): New OpenAPI route wired to the Hono app that accepts proposal ID, pagination, ordering, and optional address filter params
- **Repository** (`apps/api/src/repositories/votes/offchainNonVoters.ts`): Cross-schema query using a LEFT JOIN between `accountPower` (public schema) and `offchainVotes` (snapshot schema) to identify delegates with voting power > 0 who did not vote, with pagination and address filtering
- **Unified Drizzle client** (`apps/api/src/database/index.ts`, `apps/api/cmd/index.ts`): New `UnifiedDrizzle` type and `pgUnifiedClient` instance combining both schemas to support cross-schema queries
- **Service layer** (`apps/api/src/services/votes/offchainNonVoters.ts`): Orchestrates repository calls, converts `bigint` voting power to string, and optimizes `totalCount` by using `addresses.length` when a filter is provided
- **Zod schemas** (`apps/api/src/mappers/proposals/offchainVoters.ts`): Request/response validation schemas for the non-voters endpoint
- **Address normalization** (`apps/offchain-indexer/src/mappers/proposal.ts`, `apps/offchain-indexer/src/mappers/vote.ts`): Added `viem`'s `getAddress` transform to ensure proposal authors and voters are stored as EIP-55 checksummed addresses
- **Tests**: Integration tests for the repository (PGlite-based) and unit tests for the service layer covering pagination, ordering, address filtering, and edge cases
- **Gateway schema** (`apps/api-gateway/schema.graphql`): Regenerated with the new `offchainProposalNonVoters` query and enriched field descriptions


apps/offchain-indexer/src/mappers/vote.ts
```diff
@@ -1,8 +1,9 @@
+import { getAddress } from "viem";
 import { z } from "zod";
 
 export const rawVoteSchema = z.object({
   id: z.string(),
-  voter: z.string(),
+  voter: z.string().transform((val) => getAddress(val)),
```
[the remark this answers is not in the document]
pikonha: also `transform` generates as JSON on the `schema.graphql`, check if that's not the case here.

if not, I'd suggest this:


```suggestion
  voter: z
    .string()
    .refine(val => isAddress(val, {strict:false})
    .transform((val) => getAddress(val))
    .openapi({type: "string"}),
```

apps/offchain-indexer/src/mappers/vote.ts
pikonha, line 30: this should be done using the zod schema as mapper

pikonha, line 21: this will create a `#ref` on the openapi spec, which will then be generated as a separate ts type

```suggestion
}).openapi("OffchainVote");
```

pikonha, line 19: why would you have an empty string when there's no reason?

```suggestion
```

apps/offchain-indexer/src/mappers/proposal.ts
```diff
@@ -1,9 +1,10 @@
+import { getAddress } from "viem";
 import { z } from "zod";
 
 // "nullish().transform" is necessary here because provider return null instead of undefined
 export const rawProposalSchema = z.object({
   id: z.string(),
-  author: z.string(),
+  author: z.string().transform((val) => getAddress(val)),
```
pikonha:
```suggestion
  author: z 
            .string()
            .refine(val => isAddress(val, {strict:false})
            .transform((val) => getAddress(val))
            .openapi({type: "string"}),
```

apps/offchain-indexer/src/mappers/proposal.ts
pikonha, line 65:
```suggestion
body: z.string().optional(),

discussion: z.string().optional(),

type: z.string().optional().default("single-choice"),

start: z.number().int(),

end: z.number().int(),

state: z.string().optional().default("closed"),

created: z.number().int(),

updated: z.number().nullish(),

link: z.string().optional(),

flagged: z.boolean().optional().default(false),

scores: z.array(z.number().int()).optional().default([]),

choices: z.array(z.string()).optional().default([]),

network: z.string().optional(),

snapshot: z.number().optional(),

strategies: z
  .array(
    z.object({
      name: z.string(),
      network: z.string(),
      params: z.record(z.unknown()), // what is this for?
    })
  )
  .optional()
  .default([]),
```

apps/api/cmd/index.ts
```diff
@@ -154,6 +157,10 @@ const pgOffchainClient = drizzle(env.DATABASE_URL, {
   schema: offchainSchema,
   casing: "snake_case",
 });
+const pgUnifiedClient = drizzle(env.DATABASE_URL, {
+  schema: { ...schema, ...offchainSchema },
+  casing: "snake_case",
+});
```
pikonha: why not to use this for all queries?

apps/api/src/mappers/proposals/offchainVoters.ts
```diff
@@ -0,0 +1,18 @@
+import { z } from "@hono/zod-openapi";
+import { isAddress } from "viem";
+
+export const OffchainVoterResponseSchema = z.object({
+  voter: z.string().refine((val) => isAddress(val)),
```
pikonha:
```suggestion
  voter: z.string().refine((val) => isAddress(val, {strict: false})).transform(val => getAddress(val)),
```

apps/api/src/mappers/proposals/offchainVoters.ts
```diff
@@ -0,0 +1,18 @@
...
+  votingPower: z.string(),
+});
+
+export type OffchainVoterResponse = z.infer<typeof OffchainVoterResponseSchema>;
+
+export const OffchainVotersResponseSchema = z.object({
+  items: z.array(OffchainVoterResponseSchema),
+  totalCount: z.number(),
+});
```
pikonha:
```suggestion
export const OffchainVotersResponseSchema = z.object({
  items: z.array(OffchainVoterResponseSchema),
  totalCount: z.number(),
}).openapi("OffchainVotersResponse");
```

apps/api/src/mappers/proposals/offchainVoters.ts
```diff
@@ -0,0 +1,18 @@
+import { z } from "@hono/zod-openapi";
+import { isAddress } from "viem";
+
+export const OffchainVoterResponseSchema = z.object({
+  voter: z.string().refine((val) => isAddress(val)),
+  votingPower: z.string(),
+});
```
pikonha:
```suggestion
}).openapi("OffchainVote");
```

apps/api/src/mappers/proposals/offchainVoters.ts
```diff
@@ -0,0 +1,18 @@
...
+
+export type OffchainVoterResponse = z.infer<typeof OffchainVoterResponseSchema>;
+
+export const OffchainVotersResponseSchema = z.object({
+  items: z.array(OffchainVoterResponseSchema),
+  totalCount: z.number(),
```
pikonha:
```suggestion
  totalCount: z.number().int(),
```

apps/api/src/repositories/votes/offchainNonVoters.ts
```diff
@@ -0,0 +1,72 @@
...
+export interface OffchainNonVotersRepository {
+  getOffchainNonVoters(
+    proposalId: string,
+    skip: number,
+    limit: number,
+    orderDirection: "asc" | "desc",
+    addresses?: Address[],
+  ): Promise<{ voter: Address; votingPower: bigint }[]>;
+  getOffchainNonVotersCount(proposalId: string): Promise<number>;
+}
```
pikonha, line 24: what's the point of having the interface and the implementation on the same place? the real value comes from interfaces just with what the user needs, not all the functions implemented

apps/api/src/mappers/proposals/offchainVoters.ts
pikonha: and it should be the same from the offchain votes endpoint

apps/api/src/repositories/votes/offchainNonVoters.ts
```diff
@@ -0,0 +1,72 @@
...
+      .leftJoin(
+        offchainVotes,
+        and(
+          eq(offchainVotes.proposalId, proposalId),
+          eq(offchainVotes.voter, accountPower.accountId),
+        ),
+      )
+      .where(
+        and(
+          ...(addresses ? [inArray(accountPower.accountId, addresses)] : []),
```
pikonha, line 59: what is this used for?

apps/api/src/repositories/votes/offchainNonVoters.ts
```diff
@@ -0,0 +1,72 @@
...
+        offchainVotes,
+        and(
+          eq(offchainVotes.proposalId, proposalId),
+          eq(offchainVotes.voter, accountPower.accountId),
+        ),
+      )
+      .where(
+        and(gt(accountPower.votingPower, 0n), isNull(offchainVotes.proposalId)),
+      );
+    return countResult[0]?.count || 0;
```
pikonha, line 94:
```suggestion
    return await this.db
      .$count(accountPower.accountId)
      .from(accountPower)
      .leftJoin(
        offchainVotes,
        and(
          eq(offchainVotes.proposalId, proposalId),
          eq(offchainVotes.voter, accountPower.accountId),
        ),
      )
      .where(
        and(gt(accountPower.votingPower, 0n), isNull(offchainVotes.proposalId)),
      );
```

apps/api/src/services/votes/offchainNonVoters.ts
```diff
@@ -0,0 +1,34 @@
...
+      proposalId,
+      skip,
+      limit,
+      orderDirection,
+      addresses,
+    );
+
+    return {
+      totalCount: addresses
+        ? addresses.length
```
pikonha: what if i give an address that doesn't exist? how it would behave

pikonha (changes requested):

apps/offchain-indexer/src/mappers/vote.ts
LeonardoVieira1630: It is a zod schema

LeonardoVieira1630: The problem is that rawVoteSchema contains .transform() calls (on voter, vp, reason), which makes it a ZodEffects type. zod-openapi's .openapi() doesn't support ZodEffects directly.

LeonardoVieira1630: The reason field from Snapshot's API is string | null. The DB column (d.text().notNull().default("")) is NOT NULL, so inserting null would violate the constraint.

apps/offchain-indexer/src/mappers/proposal.ts
LeonardoVieira1630: Looking at the DB schema (offchain-schema.ts), almost every field is .notNull(). So the DB guarantees these values always exist — making them .optional() in the response schema is misleading.

LeonardoVieira1630: But i will add the ".int( )" on some fields as requested.

apps/api/src/repositories/votes/offchainNonVoters.ts
LeonardoVieira1630: The physical location (same file vs separate file) doesn't affect decoupling. The interface is already being used. To implement a stup repo in the service's tests. The last-update/index.ts on the anticapture repo also follow the same pattern.

apps/api/src/repositories/votes/offchainNonVoters.ts
LeonardoVieira1630: If addresses is provided, it adds an inArray condition to the SQL WHERE clause, restricting results to only accounts in that list. If addresses is undefined, it spreads an empty array, meaning no extra condition is added, and all addresses are returned.

apps/api/src/repositories/votes/offchainNonVoters.ts
LeonardoVieira1630: .$count() only supports SELECT count(*) FROM table WHERE ... — it doesn't accept joins. 

The isNull(offchainVotes.proposalId) condition depends on the leftJoin to detect accounts that haven't voted on the proposal (anti-join pattern). Without the join, there's no offchainVotes column in scope to check. 

Keeping the verbose approach for this reason.

apps/api/cmd/index.ts
LeonardoVieira1630: i mean, we can... but there are downsides like:
- The separate Drizzle and OffchainDrizzle types in database/index.ts act as compile-time guardrails. A repository typed as Drizzle cannot query offchainProposals as the TypeScript will refuse it. With UnifiedDrizzle everywhere, this enforcement disappears and any repo could accidentally cross schema boundaries.
- not a problem today, but spreading two schemas is a footgun if someone adds a table with the same export name to both schemas. Idkw how the collision would be handled.

apps/api/src/services/votes/offchainNonVoters.ts
LeonardoVieira1630: Flow:                                                                                    
  1. proposalExists check passes (assuming the proposal is valid)                          
  2. getOffchainNonVoters runs with inArray(accountPower.accountId, addresses) — since the 
  address doesn't exist in accountPower, the WHERE clause matches nothing                  
  3. getOffchainNonVotersCount similarly returns 0                                         
                                                                                         
  Result: 200 OK with { totalCount: 0, items: [] }

apps/offchain-indexer/src/mappers/vote.ts
pikonha: then why make it not null on the db?

pikonha: that's why you should use `parse` instead of `transform` in [here](https://github.com/blockful/anticapture/pull/1794#discussion_r3027468371)

apps/offchain-indexer/src/mappers/proposal.ts
pikonha: what's the point in having a NON NULL column and filling it with zero values? why not just make them nullable

apps/api/cmd/index.ts
pikonha: then wouldn't these bullet points already be a problem on this implementation?

apps/api/cmd/index.ts
```diff
@@ -364,6 +371,11 @@ offchainVotes(
   wrapWithTracing(new OffchainVotesService(offchainVotesRepo)),
 );
 
+const offchainNonVotersRepo = new OffchainNonVotersRepositoryImpl(
+  pgUnifiedClient,
+);
+offchainNonVoters(app, new OffchainNonVotersService(offchainNonVotersRepo));
+
```
pikonha: all of this should only be added when the dao actually supports offchain

apps/api/src/repositories/votes/offchainNonVoters.ts
```diff
@@ -0,0 +1,96 @@
+import { and, asc, desc, gt, isNull, eq, inArray, count } from "drizzle-orm";
+import { Address } from "viem";
+
+import {
+  UnifiedDrizzle,
```
pikonha, line 5:
```suggestion
  Drizzle,
```

pikonha (approved):

PR #1798, opened 2026-04-01
API healthcheck

pikonha:
## Summary
- add a /health endpoint to both API entrypoints
- make the healthcheck verify database connectivity with a lightweight query
- expose the new health query through the API gateway REST resolvers
- fix Turbo cached output declarations so workspace typecheck restores required build artifacts


apps/api/src/controllers/health/index.integration.test.ts
LeonardoVieira1630, line 1: Why dont use pglight here to simulate the dB -> controller interaction. The pgLight usage is the pattern that was implemented on the controller tests.

LeonardoVieira1630 (approved):

apps/api/src/controllers/health/index.integration.test.ts
pikonha: done

PR #1804, opened 2026-04-06
Search proposal endpoint

pikonha:
## Summary
- add `GET /proposals/search` to search on-chain proposals by partial `id` or `title`
- validate the new search query params with a dedicated request schema
- extend the proposals service and Drizzle repository to support search results and matched counts
- reuse on-chain status hydration for search results so responses stay consistent with existing proposal reads
- add unit and integration coverage for partial title and id matches


pikonha:
@claude

brunod-e (approved):

apps/api-gateway/src/resolvers/rest.ts
PedroBinotto, line 1: looking forward to the impending demise of these resolver files : )

apps/api/src/mappers/proposals/offchainProposals.ts
```diff
@@ -144,3 +144,23 @@ export const OffchainProposalsRequestSchema = z
...
+          in: "query",
+        },
+        description:
+          "Partial Snapshot proposal identifier or title to search for.",
+        example: "test",
+      }),
+    skip: paginationSkipQueryParam(),
+    limit: paginationLimitQueryParam(),
+  })
+  .openapi("OffchainProposalSearchRequest");
```
PedroBinotto, line 166: Is there any practical difference between this schema and the on-chain one? If not, couldn't these be merged together?

apps/api/src/repositories/drizzle/index.ts
```diff
@@ -23,6 +23,19 @@ import { DBProposal } from "@/mappers";
...
+
+  private buildProposalSearchWhere(query: string): SQL<unknown> {
+    const searchPattern = `%${this.escapeLikePattern(query)}%`;
+
+    return sql`(
+      ${proposalsOnchain.id} ILIKE ${searchPattern} ESCAPE '\\'
+      OR ${proposalsOnchain.title} ILIKE ${searchPattern} ESCAPE '\\'
+    )`;
+  }
+
```
PedroBinotto: these look like they'd belong in a utils file, both for semantics and for making it more testable

apps/api/src/repositories/proposals/offchainProposals.ts
```diff
@@ -1,11 +1,24 @@
...
+
+  private buildProposalSearchWhere(query: string): SQL<unknown> {
+    const searchPattern = `%${this.escapeLikePattern(query)}%`;
+
+    return sql`(
+      ${offchainProposals.id} ILIKE ${searchPattern} ESCAPE '\\'
+      OR ${offchainProposals.title} ILIKE ${searchPattern} ESCAPE '\\'
+    )`;
+  }
+
```
PedroBinotto: as per the previous comments, I'd imagine we could avoid this code duplication and make it more testable by extracting it to a utilitarian function

PedroBinotto (changes requested):
just some minor adjustments

apps/api/src/mappers/proposals/offchainProposals.ts
pikonha: wdym be `on-chain one`? it is the partial proposal `ID` or `title` that will be used as search param

PR #1808, opened 2026-04-08
Feat/cache on gateful

LeonardoVieira1630:
## Summary

Adds a Redis-backed cache layer to the API Gateway (`gateful`) to reduce upstream load and improve response latency. API controllers now emit appropriate `Cache-Control` headers, which the gateway uses as the TTL signal to populate the cache.

## Changes

- `apps/gateful/src/middlewares/cache.ts` — new cache-aside Hono middleware that reads from Redis on cache hit and writes on cache miss, driven by the upstream `Cache-Control: max-age` header; fails open on Redis errors
- `apps/gateful/src/middlewares/cache.test.ts` — unit tests covering HIT, MISS, Redis read/write errors (fail-open), non-2xx suppression, and missing `Cache-Control`
- `apps/gateful/src/cache/redis.ts` — thin Redis client factory with non-blocking connect and structured event logging
- `apps/gateful/src/index.ts` — wires the cache middleware into the gateway; Redis is optional via `REDIS_URL` env var
- `apps/gateful/src/config.ts` — adds optional `REDIS_URL` config entry
- `apps/gateful/src/shared/fan-out.ts` — refactored to return `{ data, cacheControl }` so the upstream `Cache-Control` header propagates through the fan-out layer
- `apps/gateful/src/shared/fan-out.test.ts` — integration tests for `fanOutGet` using MSW, covering header propagation and failed-upstream exclusion
- `apps/gateful/src/resolvers/daos/service.ts` + `route.ts` — forwards the aggregated `cacheControl` value to the response header
- `apps/gateful/src/resolvers/delegation/service.ts` + `route.ts` — same pattern for the delegation resolver
- `apps/api/src/controllers/**` — all API route handlers now set `Cache-Control: public, max-age=<N>` with TTLs tuned per data volatility (30 s – 1800 s)
- `apps/gateful/package.json` — adds `redis` and `msw` dependencies
- `docs/specs/2026-04-07-gateway-cache-design.md` — design spec for the cache layer


apps/api/src/controllers/account-balance/historical.ts
```diff
@@ -61,6 +61,7 @@ export function historicalBalances(
         toDate,
       );
 
+      context.header("Cache-Control", "public, max-age=60");
```
pikonha: it would be better to have this as a shared middleware that receives the seconds as arguments, that way we avoid typos on this header setting.

read [this thread](https://github.com/honojs/middleware/issues/637)

apps/gateful/src/index.ts
```diff
@@ -28,6 +30,8 @@ app.use("*", logger());
 if (config.blockfulApiToken) {
   app.use("*", bearerAuth({ token: config.blockfulApiToken }));
 }
+const redis = config.redisUrl ? createRedisClient(config.redisUrl) : undefined;
+app.use("*", cacheMiddleware(redis));
```
pikonha: why not to avoid inserting the middleware if the redis isn't there instead of having it as nullable argument?

apps/gateful/src/resolvers/delegation/service.ts
pikonha, line 1: could you explain what is being done here?

pikonha (changes requested):

pikonha (approved):

PR #1809, opened 2026-04-08
Chore/refactor hypersync

PedroBinotto:


Dockerfile.hypersync-indexer
pikonha, line 1: move this to `infra/` and add a `railway.json` file

pikonha:
# Code Review

**Status**: DRAFT · +5528 / −477 · labels: API, Indexer

## Overview

This PR introduces a new Envio/HyperIndex indexer (`@anticapture/hypersync-indexer`) as a replacement for the Ponder-based indexer, and migrates the API's Drizzle schema to match Envio's default naming convention. Changes span three areas:

1. **New app** `apps/hypersync-indexer/` — Envio project for ENS token + governor with handlers for transfers, delegations, voting, and supply metrics, plus a root `Dockerfile.hypersync-indexer`.
2. **API schema rewrite** (`apps/api/src/database/schema.ts`) — tables renamed `snake_case` → `PascalCase` (`account_balance` → `AccountBalance`), columns renamed to `camelCase`, `casing: \"snake_case\"` dropped from Drizzle clients, composite primary keys replaced with synthetic `id` text PKs, `DATABASE_URL` no longer injects `search_path=anticapture`.
3. **Test sweep** — every unit/integration test updated for the new schema shape.

---

## Must-fix

- **Empty `ts-node` file at repo root** (`diff --git a/ts-node b/ts-node`, 0 bytes). Almost certainly an accidental artifact of `npm install` with a typo / redirected output. Delete.
- **Empty PR description on a 5.5k-line structural migration.** Please add: motivation, rollout/migration plan for prod DBs, coexistence story with the existing Ponder indexer, rollback procedure. `DRAFT` is fine, but reviewers need this.
- **Lost DB-level uniqueness constraints.** The old schema had composite PKs: `AccountBalance(accountId, tokenId)`, `AccountPower(accountId)`, `VotingPowerHistory(transactionHash, accountId, logIndex)`, `BalanceHistory(...)`. They're now synthetic `id text primary key` columns with no unique index on the natural key. Integrity is now guaranteed only by the indexer's id-construction convention (`\${accountId}-\${tokenId}`, etc.). Add unique indexes on the natural keys — this is the kind of thing you want enforced by Postgres, not application code.
- **`DATABASE_URL` search_path override removed** (`apps/api/src/env.ts`) and `casing: \"snake_case\"` dropped (`cmd/index.ts`, `cmd/aave.ts`). Confirm Envio writes to whatever schema the API reads from (previously the `anticapture` schema was forced). If Envio writes to `public` by default, the API now silently reads from `public` too. Document the expected DB layout and whether existing `anticapture`-schema data is being dropped.
- **`apps/hypersync-indexer/generated/package.json` is checked in** but the surrounding `.gitignore` ignores `/generated/*` with an exception for this one file. `envio codegen` regenerates the whole directory; pinning `package.json` will produce diff churn and drift vs. the installed Envio version. Either gitignore the whole folder and have CI run `envio codegen && pnpm install` in `installer`, or commit the full `generated/` (not recommended). The current hybrid is the worst option.

## Should-fix

- **Dockerfile.hypersync-indexer**
  - `envio` is in `devDependencies`, but the runner sets `NODE_ENV=production` and the entrypoint runs `npm run start` → `envio start`. If `pnpm install` ever skips devDeps, the binary disappears. Move `envio` to `dependencies` or set `--prod=false` explicitly.
  - `npm install --prefix generated --legacy-peer-deps` in a pnpm monorepo undermines lockfile determinism — this is why committing `generated/package.json` is fragile. Prefer `pnpm install` at the generated dir, or regenerate at build time and install once.
  - Installs `curl` and `ts-node` globally in both `installer` and `runner` stages — collapse into the stage that needs it.
  - No `USER` directive (runs as root). No `HEALTHCHECK`. Standard for a production indexer.
- **`handleTransaction` early-returns when no flag matches** (`shared.ts`). Correct for intent (only persisting flagged txs), but the function name suggests unconditional persistence. Rename or add a one-line comment.
- **`storeDailyBucket.average` uses integer `bigint` division** (`shared.ts`), which will drift for large counts. Same pattern as before, so not a regression, but worth noting.
- **Enum string mapping duplicated** between `MetricTypesEnum` (TS) and `metricType: MetricType_t` (generated from `schema.graphql`). Any drift is silent. Consider generating one from the other or adding an exhaustive-check.
- **`_delegatorBalance` leading underscore** (`delegation.ts`) — the variable is read immediately after, so the underscore is misleading. Drop it.
- **`ensureAccountExists` + `ensureAccountsExist`** duplicate each other; keep only the batched version.
- **`delegation.ts` delegateChanged** accumulates `existingDelegation.delegatedValue + delegatedValue`. `DelegateChanged` per delegator-per-tx is normally unique, but if the id ever collides (re-org replay, same tx replayed), this double-counts. Worth a comment on intent.
- **No handler tests**: the new indexer has ~1.5k LOC of business logic and zero tests. At minimum, unit tests for `handleTransaction`, `storeDailyBucket`, and `delegateChanged` would catch future regressions. The PR updates API tests exhaustively — please extend that rigor to the new package.

## Nits

- `eslint.config.mjs` — the comment says \"const + type with same name pattern\"; point to the file in `src/lib/` where that pattern lives so future readers don't re-litigate.
- `apps/hypersync-indexer/tsconfig.json` includes `generated/**/*.ts` with `allowImportingTsExtensions: true` + `noEmit` — only meaningful at typecheck time. Worth documenting for newcomers.
- Root `package.json` adds `hypersync-indexer` turbo script — consistent with existing `indexer`/`api` commands, good.

## Verification checklist

- [ ] Start Envio locally against a fresh Postgres; confirm API reads the same schema Envio writes (search_path alignment).
- [ ] Run `pnpm api typecheck && pnpm api test` after schema cutover.
- [ ] Spot-check that `row.Transfer` / `row.Delegation` join aliases in `voting-power/general.ts` and `nouns.ts` actually match Drizzle's relation inference for the new PascalCase table names — this typechecks but could fail at runtime.
- [ ] Confirm rollback plan: can we re-run the old Ponder indexer against the same DB if the Envio cutover fails? If not, note that explicitly.

PR #1812, opened 2026-04-13
Feat/circuit breaker on gateful

LeonardoVieira1630:
## Summary

Add a circuit breaker with exponential backoff to the Gateful API gateway, protecting the system from cascading failures when upstream DAO APIs become unavailable. When a DAO backend exceeds a configurable failure threshold, its circuit opens and requests are fast-failed with a 503 response until a cooldown-based probe succeeds.

## Changes

- **New `CircuitBreaker` class** (`shared/circuit-breaker.ts`): implements the CLOSED -> OPEN -> HALF_OPEN state machine with configurable failure threshold, cooldown, exponential backoff, and max cooldown cap
- **New `CircuitBreakerRegistry`** (`shared/circuit-breaker-registry.ts`): lazily creates and manages per-DAO circuit breaker instances
- **Config**: added `CIRCUIT_BREAKER_FAILURE_THRESHOLD`, `CIRCUIT_BREAKER_COOLDOWN_MS`, and `CIRCUIT_BREAKER_MAX_COOLDOWN_MS` env vars with sensible defaults
- **Health endpoint**: now returns per-DAO circuit breaker states (state + next retry time) in the `/health` response
- **Proxy route**: wraps upstream calls in the circuit breaker; 5xx responses from upstreams count as failures
- **Fan-out utility**: routes each DAO fetch through its circuit breaker; downgrades `CircuitOpenError` logs from error to warn
- **Services** (`DaosService`, `DelegationService`): updated constructors to accept and pass through the registry
- **Global error handler**: catches `CircuitOpenError` and returns a 503 JSON response
- **Tests**: added unit tests for the circuit breaker state transitions and updated existing tests to supply the registry

Uploading Screen Recording 2026-04-13 at 15.39.47.mov…


pikonha:
## Code Review

The PR introduces a well-structured circuit breaker pattern with proper state machine transitions, exponential backoff, per-DAO isolation, health endpoint visibility, and good test coverage. The integration across proxy, fan-out, and service layers is consistent.

**1 issue found:**

### #1 Client errors (4xx) incorrectly trip the circuit breaker in the fan-out path

In `apps/gateful/src/shared/fan-out.ts:26`, `if (!res.ok)` throws for any non-2xx response — including 4xx client errors (400, 404, 422) — inside `registry.get(dao).execute(...)`, so it counts toward the circuit breaker's failure threshold. Five 4xx responses in a row will open the circuit and block all requests to that DAO for the entire cooldown period, even though the upstream is healthy.

The proxy route (`apps/gateful/src/proxy/route.ts:50`) correctly only counts `res.status >= 500`. The fan-out path should apply the same filter — only throw for 5xx inside the execute callback, and handle 4xx outside it.

pikonha (approved):

PR #1814, opened 2026-04-13
Kubb client

pikonha:
## Summary

- Add [Kubb](https://kubb.dev/) to generate TypeScript API client and React Query hooks directly from the OpenAPI spec
- Add `anticapture-client` package with auto-generated REST hooks, suspense support, and infinite query hooks
- Add Gateful proxy route on the dashboard to handle API gateway authentication


apps/dashboard/app/api/gateful/[...path]/route.ts
Comment (pikonha) [inline apps/dashboard/app/api/gateful/[...path]/route.ts:1]: this is used to avoid leaking the auth token by making a request to the Next API which then injects that private variables to the request

.github/workflows/api-contract-updates.yaml
pikonha, line 1: this forces the `schema.graphql` and kubb generated files to be updated whenever there's a change on the api's mappers and controllers

apps/gateful/openapi/gateful.json
brunod-e, line 1: do we really need to keep a 6.4k lines file in our project? it couldnt be generated on build or something?

packages/anticapture-client/kubb.config.ts
```diff
@@ -0,0 +1,71 @@
...
+        {
+          type: "operationId",
+          pattern: "dao",
+        },
+      ],
+      mutation: false,
+      transformers: {
+        name: (name, type) =>
+          name === "dao" && type === "function" ? "getDao" : name, // rename "dao" operation ID to getDao
+      },
```
brunod-e, line 55: you are transforming dao at same time you exclude it

packages/anticapture-client/kubb.config.ts
```diff
@@ -0,0 +1,71 @@
...
+      client: {
+        importPath: "../src/client",
+      },
+      exclude: [
+        {
+          type: "operationId",
+          pattern: "dao",
+        },
+      ],
+      mutation: false,
```
brunod-e, line 51: we dont have any mutations in our api?

packages/anticapture-client/kubb.config.ts
```diff
@@ -0,0 +1,71 @@
...
+      transformers: {
+        name: (name, type) =>
+          name === "dao" && type === "function" ? "getDao" : name, // rename "dao" operation ID to getDao
+      },
+      suspense: {},
```
brunod-e, line 56: why this is empty? remove or enable it

brunod-e (changes requested):

packages/anticapture-client/kubb.config.ts
pikonha: nope

apps/gateful/openapi/gateful.json
pikonha: you mean the openapi spec? if so, it is used to ensure the gateway has been run locally, as well as being the source for generating the SDKs. if we do not have it versioned, we gonna have to deploye the api and gateway to dev everytime so the codegen is updated

packages/anticapture-client/kubb.config.ts
pikonha: [the empty object is enalbing it](https://kubb.dev/kubb/plugins/plugin-react-query#suspense)

packages/anticapture-client/kubb.config.ts
pikonha: it is needed otherwise it genereates useDao query

PR #1819, opened 2026-04-14
Governance Relayer

LeonardoVieira1630:
Implements a standalone Hono service (apps/relayer/) that allows users to vote and delegate on governance proposals without paying gas. The relayer validates EIP-712 signatures, checks eligibility and on-chain state, enforces rate limits, and submits transactions via a funded wallet.

- Pluggable signer interface (local private key for v1, KMS-ready)
- Supports castVoteBySig and delegateBySig for all supported DAOs
- In-memory sliding window rate limiter (per-address hourly + daily)
- Zod OpenAPI schemas with /health, /relay/vote, /relay/delegate routes
- 29 tests across 7 files (unit + integration with MSW)


apps/relayer/src/index.ts
```diff
@@ -0,0 +1,169 @@
...
+import { SignatureVerifier } from "@/services/guards/signature-verifier";
+import { createLocalSigner } from "@/signer/local-signer";
+
+const logger = pino({ name: "relayer" });
+
+const allChains = Object.values(chains);
+
+async function main() {
+  const chain = extractChain({
+    chains: allChains,
```
pikonha: change this to `mainnet`

apps/relayer/src/env.ts
```diff
@@ -0,0 +1,47 @@
...
+
+dotenv.config();
+
+const envSchema = z.object({
+  REDIS_URL: z.url(),
+
+  DAO_NAME: z.string().min(1),
+
+  RPC_URL: z.string().url(),
+  CHAIN_ID: z.coerce.number(),
```
pikonha:
```suggestion
  CHAIN_ID: z.coerce.number().int().openapi({type:"integer"}),
```

apps/relayer/src/index.ts
```diff
@@ -0,0 +1,169 @@
...
+    id: env.CHAIN_ID as (typeof allChains)[number]["id"],
```
pikonha: i suppose you have to cast here because `CHAIN_ID` from env with `coerce` makes it optional, which is not really the case. change its type and remove this manual cast


```suggestion
    id: env.CHAIN_ID,
```

apps/relayer/src/signer/local-signer.ts
```diff
@@ -0,0 +1,42 @@
...
+      to: Address;
+      data: Hex;
+      value?: bigint;
+    }): Promise<Hash> {
+      return client.sendTransaction({
+        to: tx.to,
+        data: tx.data,
+        value: tx.value ?? 0n,
+      });
+    },
```
pikonha, line 38:
```suggestion
    async sendTransaction(tx: {
      to: Address;
      data: Hex;
      value: bigint = 0n;
    }): Promise<Hash> {
      return client.sendTransaction(tx);
    },
```

apps/relayer/src/signer/types.ts
```diff
@@ -0,0 +1,16 @@
+import { Address, Hash, Hex } from "viem";
+
+export interface RelayerSigner {
+  /** The Ethereum address of this signer */
+  readonly address: Address;
```
pikonha: make it a function

apps/relayer/src/env.ts
```diff
@@ -0,0 +1,47 @@
...
+    .transform((v) => getAddress(v)),
+  TOKEN_ADDRESS: z
+    .string()
+    .refine(isAddress, "Invalid Ethereum address")
+    .transform((v) => getAddress(v)),
+
+  RELAYER_PRIVATE_KEY: z
+    .string()
+    .refine(isHex, "Invalid hex string")
+    .transform((v) => v as `0x${string}`),
```
pikonha, line 27:
```suggestion,
```

apps/relayer/src/env.ts
```diff
@@ -0,0 +1,47 @@
...
+
+  MIN_VOTING_POWER: z.string().default("0"),
+
+  MAX_RELAY_PER_ADDRESS_PER_DAY: z.coerce.number().default(3),
```
pikonha:
```suggestion
  MAX_RELAY_PER_ADDRESS_PER_DAY: z.coerce.number().integer().optional().default(3),
```

apps/relayer/src/env.ts
```diff
@@ -0,0 +1,47 @@
...
+
+  MIN_BALANCE_WEI: z.string().default("100000000000000000"),
```
pikonha: NOUNS have 0 decimals, the others have 18

```suggestion
  MIN_BALANCE_ETHER: z.string().default("0.1").transform(val => parseUnits(val, decimals)),
```

apps/relayer/src/env.ts
```diff
@@ -0,0 +1,47 @@
...
+});
+
+const _env = envSchema.safeParse(process.env);
+
+if (!_env.success) {
+  console.error("Invalid environment variables", _env.error.issues);
+  throw new Error("Invalid environment variables");
+}
+
+export const env = _env.data;
```
pikonha:
```suggestion
export const env = envSchema.parse(process.env);
```

apps/relayer/src/env.ts
```diff
@@ -0,0 +1,47 @@
...
+    .refine(isHex, "Invalid hex string")
+    .transform((v) => v as `0x${string}`),
+
+  MIN_VOTING_POWER: z.string().default("0"),
+
+  MAX_RELAY_PER_ADDRESS_PER_DAY: z.coerce.number().default(3),
+
+  MIN_BALANCE_WEI: z.string().default("100000000000000000"),
+
+  BLOCKFUL_API_TOKEN: z.string().optional(),
```
pikonha: this should be handled by the gateful

apps/relayer/src/env.ts
```diff
@@ -0,0 +1,47 @@
...
+
+  PORT: z.coerce.number().default(4001),
```
pikonha:
```suggestion
  PORT: z.coerce.number().default(3002),
```

apps/relayer/src/env.ts
```diff
@@ -0,0 +1,47 @@
...
+import { z } from "zod";
+
+dotenv.config();
+
+const envSchema = z.object({
+  REDIS_URL: z.url(),
+
+  DAO_NAME: z.string().min(1),
+
+  RPC_URL: z.string().url(),
```
pikonha:
```suggestion
  RPC_URL: z.url(),
```

apps/relayer/src/index.ts
```diff
@@ -0,0 +1,169 @@
...
+
+  const rateLimitStorage = new RedisRateLimitStorage(redis);
+
+  const rateLimiter = new RateLimiter(rateLimitStorage, {
+    daoName: env.DAO_NAME,
+    governorAddress: governorAddress,
+    maxPerAddressPerDay: env.MAX_RELAY_PER_ADDRESS_PER_DAY,
+  });
+
+  const minBalanceWei = BigInt(env.MIN_BALANCE_WEI);
```
pikonha:
```suggestion
```

apps/relayer/src/index.ts
```diff
@@ -0,0 +1,169 @@
...
+
+  const relayService = new RelayService(
+    signer,
+    signatureVerifier,
+    chainState,
+    rateLimiter,
+    publicClient,
+    minBalanceWei,
```
pikonha:
```suggestion
    env.MIN_BALANCE_ETHER,
```

apps/relayer/src/index.ts
```diff
@@ -0,0 +1,169 @@
...
+          400,
+        );
+      }
+    },
+  });
+
+  app.use(cors({ origin: "*" }));
+  if (env.BLOCKFUL_API_TOKEN) {
+    app.use(bearerAuth({ token: env.BLOCKFUL_API_TOKEN }));
+  }
```
pikonha: this will be a private service

```suggestion
```

apps/relayer/src/env.ts
```diff
@@ -0,0 +1,47 @@
+import dotenv from "dotenv";
+import { getAddress, isAddress, isHex } from "viem";
+import { z } from "zod";
+
+dotenv.config();
+
+const envSchema = z.object({
```
pikonha, line 7: add the OTEL exporter env. in fact, add all the observability to this service

apps/relayer/src/controllers/relay-vote.ts
```diff
@@ -0,0 +1,73 @@
...
+          content: {
+            "application/json": { schema: ErrorResponseSchema },
+          },
+        },
+      },
+    }),
+    async (c) => {
+      const body = c.req.valid("json");
+      const result = await relayService.relayVote({
+        proposalId: BigInt(body.proposalId),
```
pikonha: I'm not entirely sure this works

apps/relayer/src/services/relay.ts
```diff
@@ -0,0 +1,167 @@
...
+export interface RelayResult {
+  hash: Hash;
+  signer: Address;
+}
+
+export class RelayService {
+  constructor(
+    private signer: RelayerSigner,
+    private signatureVerifier: ISignatureVerifier,
+    private chainState: IChainStateService,
```
pikonha, line 47: this breaks interface segregation

apps/relayer/src/services/relay.ts
```diff
@@ -0,0 +1,167 @@
...
+    private publicClient: ChainReader,
+    private minBalanceWei: bigint,
+    private minVotingPower: bigint,
+    private governorAddress: Address,
+    private tokenAddress: Address,
+  ) {}
+
+  async relayVote(params: RelayVoteParams): Promise<RelayResult> {
+    // 0. Check relayer has enough balance
+    await this.assertSufficientBalance();
```
pikonha: this could be handled by the simulateContract reducing by 1 the number of RPC calls

apps/relayer/src/services/relay.ts
```diff
@@ -0,0 +1,167 @@
...
+      throw Errors.PROPOSAL_NOT_ACTIVE();
+    }
+
+    const alreadyVoted = await this.chainState.hasVoted(
+      params.proposalId,
+      voter,
+    );
+    if (alreadyVoted) {
+      throw Errors.ALREADY_VOTED();
+    }
```
pikonha: this could be handled by the simulateContract reducing by 1 the number of RPC calls

apps/relayer/src/schemas/relay-delegate.ts
```diff
@@ -0,0 +1,29 @@
...
+  .object({
+    delegatee: AddressSchema,
+    nonce: z.string().openapi({
+      description: "Delegation nonce as decimal string",
+      example: "0",
+    }),
+    expiry: z.string().openapi({
+      description: "Signature expiry timestamp as decimal string",
+      example: "1718000000",
+    }),
```
pikonha: where does this comes from?

apps/relayer/src/schemas/relay-delegate.ts
```diff
@@ -0,0 +1,29 @@
...
+
+import { AddressSchema, Bytes32Schema, TxHashSchema } from "./evm-primitives";
+
+export const RelayDelegateRequestSchema = z
+  .object({
+    delegatee: AddressSchema,
+    nonce: z.string().openapi({
+      description: "Delegation nonce as decimal string",
+      example: "0",
+    }),
```
pikonha: check whether this is really required

apps/relayer/src/schemas/relay-delegate.ts
pikonha: if its really required, shouldn't it be for voting as well?

pikonha (changes requested):

pikonha (commented):
## Code Review

### Architecture & Design

**Strengths:**
- Clean layered separation: `controller → service → signer/chain/guards/repository`
- Pluggable interfaces on every dependency (`IChainStateService`, `IRateLimiter`, `ISignatureVerifier`, `RelayerSigner`) — KMS upgrade path is clear
- `RelayError` factory pattern avoids shared mutable instances across requests
- `ChainReader` narrow interface enables lightweight test doubles without mocking all of `PublicClient`
- Zod env validation at startup with useful error messages

**Concerns:**
- **Race condition on rate limit** (`relay.ts` step 4): Rate limiter is checked *after* on-chain validation but *before* submission. If two concurrent requests from the same address pass on-chain checks simultaneously, both can consume a rate-limit slot and still submit. Worth documenting as a known limitation.
- **`version: "1"` hardcoded** (`index.ts`, `app.ts`): EIP-712 domain `version` is always `"1"`. Some governance tokens use a different version string — this will silently produce wrong signature recovery and `INVALID_SIGNATURE` errors for non-standard contracts. Consider fetching from the contract or making it configurable via env.

---

### Security

- **`cors({ origin: "*" })`**: Wildcard CORS on a service that handles financial transactions is risky, especially since bearer auth is optional. If `BLOCKFUL_API_TOKEN` is unset, any origin can relay transactions. Consider restricting origins or requiring the token unconditionally in production.
- **EIP-712 recovery never fails**: `recoverTypedDataAddress` always returns *an* address — a garbled/wrong-domain signature returns a wrong address rather than throwing. A caller who signs with the wrong domain gets `INSUFFICIENT_VOTING_POWER` instead of `INVALID_SIGNATURE`, which is confusing to debug. Low severity but worth noting.

---

### Correctness

- **`getVotes` vs `getPastVotes`** (`chain-state.ts`, `getVotingPower`): Uses `getVotes(address)` which returns *current* voting power. OpenZeppelin Governor uses `getPastVotes(address, proposalSnapshot)` — the power at the snapshot block — for voting eligibility. A user who acquires tokens *after* a snapshot will pass this eligibility check but still fail on-chain (or lose tokens and be denied despite having eligible snapshot votes). This is a meaningful correctness bug for the `relayVote` path.

- **Rate limit slot consumed before tx submission**: If `signer.sendTransaction` throws, the rate limit slot is already spent (step 4 precedes step 5 in `relayVote`/`relayDelegation`). Callers hitting transient RPC errors lose their daily quota. Consider whether to decrement on failure or document this behavior.

- **`assertSufficientBalance` per request**: Fetches relayer ETH balance on every relay call. Fine for v1 but under load this adds an extra RPC round-trip per request.

---

### Code Quality

- **`as 400` cast in error handler** — this is the most important fix:
  ```ts
  // Current — always sends HTTP 400, even for 429/503:
  return c.json({ error: err.message, code: err.code }, err.status as 400);
  
  // Should be:
  import type { ContentfulStatusCode } from "hono/utils/http-status";
  return c.json({ error: err.message, code: err.code }, err.status as ContentfulStatusCode);
  ```
  This cast silences a real type error and causes every `RelayError` (including `RATE_LIMITED` → 429 and `RELAYER_LOW_BALANCE` → 503) to return HTTP 400 instead. This bug is present in both `index.ts` and `e2e/helpers/app.ts`.

- **9-argument constructor on `RelayService`**: High arity makes instantiation fragile. A config bag `{ minBalanceWei, minVotingPower, governorAddress, tokenAddress }` would read cleaner and is less likely to produce silent argument ordering bugs.

- **`vrsToSignature` manual hex concatenation** (`signature-verifier.ts`): Manually slicing and concatenating hex strings is fragile. viem exposes `serializeSignature({ r, s, v })` which is safer and more idiomatic.

- **`dotenv.config()` at module level in `env.ts`**: Calling this as a side-effect of an import means any test that imports from `@/env` attempts to read `.env` from whatever the cwd is at test time. The e2e tests work around this via `dotenv --` in the script, but unit tests don't have that protection. Consider passing `process.env` explicitly or guarding with `NODE_ENV !== "test"`.

---

### Testing

**Strong coverage** — the unit tests with clean fakes (no mocking frameworks) and the e2e tests with real Anvil fork are well done.

**Gaps:**
- No test verifies that 429/503 `RelayError`s actually return the correct HTTP status — the `as 400` bug goes undetected because all tested error cases happen to return 400.
- No test for the `/health` endpoint.
- `relay-delegate.test.ts` has no rate-limit case (covered in unit tests but not e2e).
- Missing concurrent-request test to exercise the rate-limit TOCTOU scenario.

---

### Priority Summary

| # | Severity | Issue |
|---|----------|-------|
| 1 | **High** | `as 400` cast breaks 429/503 HTTP status codes |
| 2 | **Medium** | `getVotes` instead of `getPastVotes` — wrong snapshot semantics for vote eligibility |
| 3 | **Medium** | Hardcoded `version: "1"` in EIP-712 domain breaks non-standard tokens |
| 4 | **Low** | Wildcard CORS with optional auth |
| 5 | **Low** | Rate-limit slot consumed before tx submission (document or fix) |

Overall this is a solid, well-structured service. The layering and test coverage are genuinely good. Fix the status code cast and the `getPastVotes` issue before merging.

apps/relayer/src/schemas/relay-delegate.ts
LeonardoVieira1630: to call delegateBySig function, we need to pass the expiry as a parameter. So, it is needed. The voting function does not require this.

apps/relayer/src/schemas/relay-delegate.ts
LeonardoVieira1630: to call delegateBySig function, we need to pass the nonce as a parameter. So, it is needed.

apps/relayer/e2e/relay-delegate.test.ts
```diff
@@ -0,0 +1,193 @@
...
+      hash: responseBody.transactionHash,
+    });
+    const currentDelegate = await testClient.readContract({
+      address: TOKEN_ADDRESS,
+      abi: erc20VotesAbi,
+      functionName: "delegates",
+      args: [TEST_USER_ADDRESS],
+    });
+    expect(currentDelegate.toLowerCase()).toBe(DELEGATEE_ADDRESS.toLowerCase());
+  });
```
pikonha, line 136: lit 🔥🔥🔥🔥🔥🔥🔥🔥🔥

pikonha (approved):

PR #1820, opened 2026-04-14
fix: proposals overlapping in distribution chart

PedroBinotto:


apps/dashboard/features/token-distribution/hooks/useChartMetrics.tsx
```diff
@@ -209,14 +209,23 @@ export const useChartMetrics = ({
...
+        const existing = result[timestamp];
+        const existingCount =
+          typeof existing?.PROPOSALS_GOVERNANCE === "number"
+            ? existing.PROPOSALS_GOVERNANCE
+            : 0;
+        const existingText =
+          typeof existing?.PROPOSALS_GOVERNANCE_TEXT === "string" &&
+          existing.PROPOSALS_GOVERNANCE_TEXT.length > 0
+            ? existing.PROPOSALS_GOVERNANCE_TEXT + ", "
+            : "";
```
pikonha: could you explain this?

PedroBinotto: This used to generate the text content for a point in the graph where more than one proposal was to be displayed; 

ex.: If proposals `A` and `B` point to the same block/timestamp, this would generate a proposal count of 2 and a display text of `A, B`;

This behaviour was slightly tweaked so that the description text just shows `A + n` for more than one proposal

pikonha (approved):

PR #1825, opened 2026-04-16
Backend Preview

PedroBinotto:


apps/api/cmd/seed-ci.ts
```diff
@@ -0,0 +1,174 @@
...
+
+import * as schema from "@/database/schema";
+import { env } from "@/env";
+import { logger } from "@/logger";
+
+export function isCi(): boolean {
+  return !["dev", "production"].includes(
+    process.env.RAILWAY_ENVIRONMENT_NAME || "dev",
+  );
+}
```
PedroBinotto: Unfortunately this keeps the implementation coupled to the Railway environment for CI; as we're relying on the PR environments for this deployment (which don't allow for separate environment configuration), I don't think there'll be a way to work around this

pikonha (commented):
## Code Review

**Overview:** This PR sets up CI-specific backend infrastructure for Railway preview environments. It adds automated DB migration + data seeding when the API starts in a non-dev/non-production Railway environment. It also replaces Hono's built-in logger with a custom structured logger, and disables the indexer in PR environments with a NOOP.

---

### What's Good

- **Structured logging** in `aave.ts`/`index.ts` is a clean improvement — captures method, path, status, and duration in one place and integrates with the project's pino logger.
- **Indexer NOOP in PR envs** (`infra/indexer/railway.json`) is the right call — no point running the indexer in preview.
- **Schema default fix** (`0n` → `sql`0``) — necessary for `drizzle-seed` compatibility, and it's the more correct way to express a DB-level default.
- **Batched inserts** with the `i += 500` chunking pattern avoids overwhelming the DB.

---

### Issues

**1. `isCi()` naming is misleading — and the logic is fragile**
`apps/api/cmd/seed-ci.ts:121`
```ts
export function isCi(): boolean {
  return !["dev", "production"].includes(
    process.env.RAILWAY_ENVIRONMENT_NAME || "dev",
  );
}
```
This isn't "is CI" — it's "is a Railway preview environment". If someone runs locally with `RAILWAY_ENVIRONMENT_NAME=staging`, they'll trigger truncation and seeding of whatever DB is in their `DATABASE_URL`. Rename to `isRailwayPreviewEnv()` and consider adding a guard that also checks for an explicit `CI=true` env var or a dedicated `SEED_ON_START=true` flag.

---

**2. Double drizzle client in `runCiSeed` — the passed-in `pgClient` is barely used**
`apps/api/cmd/seed-ci.ts:126`
```ts
export async function runCiSeed(pgClient: ReturnType<typeof drizzle>) {
  await migrate(pgClient, { migrationsFolder: "./drizzle" });
  const ciClient = drizzle(env.DATABASE_URL, { schema, casing: "snake_case" });
  // ... uses ciClient for everything else
}
```
`pgClient` is only used for `migrate()` but then a new `ciClient` is immediately created from `env.DATABASE_URL`. This is confusing — either pass the connection string/pool and create one client, or use the passed client throughout. Also, `ciClient` uses `env.DATABASE_URL` directly without a `search_path` override, while the migration itself runs via `pgClient` (which may or may not have `search_path=anticapture` set). This is the most likely source of subtle schema routing bugs.

---

**3. Schema mismatch risk between migration runner and ciClient**
`drizzle.config.ts` sets `search_path=anticapture` via URL params for `drizzle-kit`. The migration SQL creates tables without schema prefix (they land in whatever schema is the active `search_path`). The TRUNCATE in `runCiSeed` explicitly uses `anticapture.*` prefixes. But `ciClient = drizzle(env.DATABASE_URL, ...)` — if `DATABASE_URL` doesn't include the `?options=-c search_path%3Danticapture` parameter, inserts go to `public` while the TRUNCATE targets `anticapture`, causing a disconnect.

This needs either:
- `ciClient = drizzle(env.DATABASE_URL + "?options=-c%20search_path%3Danticapture", ...)`, or
- Confirmation that `DATABASE_URL` in the CI Railway environment already sets the search path at the DB level.

---

**4. `drizzle-seed` should be a `devDependency`**
`apps/api/package.json:32`
`drizzle-seed` is only used for CI seeding and ships into the production bundle unnecessarily. Move it to `devDependencies`.

---

**5. Seed blocks server startup with no timeout**
`apps/api/cmd/index.ts:100`
```ts
if (CI) await runCiSeed(pgClient);
serve({ ... });
```
If the seed fails (DB not ready, migration error, constraint violation), the server never starts and Railway will fail the deploy. There's no timeout or graceful degradation. At minimum, wrap in try/catch and log the error — either proceed anyway or fail fast with a clear message.

---

**6. `accountBalance` + `votesOnchain` manual insert pattern is fragile**
`seed-ci.ts:209/248`
The manual inserts for `accountBalance` and `votesOnchain` are done first, then `count: 0` is passed to `drizzle-seed` to skip them. This is correct but easy to break — if someone adds a table later and forgets to handle it, seed might fail on FK constraints. A comment here would help future maintainers understand the pattern.

---

### Minor / Nits

- `const CI = isCi()` at module top-level in both `aave.ts` and `index.ts` — the name `CI` collides with the conventional `CI=true` env var set by most CI systems, which could cause confusion.
- Generated addresses `0x${String(i).padStart(40, "0")}` are not checksummed — fine for test data but worth noting if any code path validates EIP-55 checksum.
- `ws` upgraded from `7.5.10` → `8.18.0` across many transitive deps — worth verifying there are no breaking changes for websocket-dependent packages in the stack.

PedroBinotto:
> ### 3. Schema mismatch risk between migration runner and ciClient
>
> `drizzle.config.ts` sets `search_path=anticapture` via URL params for drizzle-kit. The migration SQL creates tables without schema prefix (they land in whatever schema is the active search_path). The `TRUNCATE` in `runCiSeed` explicitly uses `anticapture.*` prefixes. But `ciClient = drizzle(env.DATABASE_URL, ...)` — if `DATABASE_URL` doesn't include the `?options=-c search_path%3Danticapture` parameter, inserts go to public while the `TRUNCATE` targets anticapture, causing a disconnect.
>
> This needs either:
>
> - `ciClient = drizzle(env.DATABASE_URL + "?options=-c%20search_path%3Danticapture", ...)`, or
> - Confirmation that `DATABASE_URL` in the CI Railway environment already sets the search path at the DB level.

env.DATABASE_URL already has the search path baked in — the env.ts transform appends `?options=-c%20search_path%3Danticapture` unconditionally to whatever raw `DATABASE_URL` is in the environment.

So any client created from `env.DATABASE_URL` (including `pgClient` in the entrypoints, and now inside `runCiSeed` after the fix) already routes to anticapture.

This was actually the root cause of an earlier invalid command-line argument error — when code tried to append the search path a second time on top of what `env.ts` had already added.

apps/api/src/env.ts
```diff
@@ -10,6 +10,12 @@ const envSchema = z.object({
   DATABASE_URL: z
     .string()
     .transform((val) => `${val}?options=-c%20search_path%3Danticapture`),
+  DATABASE_PUBLIC_URL: z
+    .string()
+    .optional()
+    .transform((val) =>
+      val ? `${val}?options=-c%20search_path%3Danticapture` : undefined,
+    ),
```
pikonha:
```suggestion
  DATABASE_PUBLIC_URL: z
    .string()
    .transform((val) =>
      `${val}?options=-c%20search_path%3Danticapture`,
    )
    .optional(),
```

apps/api/cmd/index.ts
```diff
@@ -174,7 +182,16 @@ if (!daoClient) {
   throw new Error(`Client not found for DAO ${env.DAO_ID}`);
 }
 
-const pgClient = drizzle(env.DATABASE_URL, { schema, casing: "snake_case" });
+if (CI && !env.DATABASE_PUBLIC_URL) {
+  const message = "DATABASE_PUBLIC_URL is required in CI environments";
+  logger.error(message);
+  throw new Error(message);
+}
```
pikonha: couldn't this check be moved to the env file? and if its a preview env it could just override the `DATABASE_URL` so no other change would be needed

apps/api/package.json
```diff
@@ -16,7 +16,8 @@
     "lint": "eslint src",
     "lint:fix": "eslint src --fix",
     "dev": "tsx watch cmd/index.ts",
-    "dev:rw": "sh -c 'dao_id=\"$1\"; exec railway run -e dev -s \"${dao_id}-api\" -- pnpm dev' --"
+    "dev:rw": "sh -c 'dao_id=\"$1\"; exec railway run -e dev -s \"${dao_id}-api\" -- pnpm dev' --",
+    "db:migrate": "drizzle-kit migrate"
```
pikonha:
```suggestion
```

pikonha (approved):

PR #1835, opened 2026-04-23
Feat/observability package GitHub

LeonardoVieira1630:
## Summary

Prepares `@anticapture/observability` for publishing to npm and restructures its OpenTelemetry dependency so consumers share a single `@opentelemetry/api` instance. Bumps the package to `1.0.1` and wires each consuming app to depend on `@opentelemetry/api` directly to satisfy the new peer dependency.

## Changes

- `packages/observability/package.json`: bump version to `1.0.1`; add `files: ["dist"]`, `publishConfig.access: public`, and `repository` metadata for npm publishing.
- `packages/observability/package.json`: move `@opentelemetry/api` from `dependencies` to `devDependencies` + `peerDependencies` to avoid duplicate tracer/meter instances in consumers.
- `apps/address-enrichment`, `apps/api-gateway`, `apps/gateful`, `apps/offchain-indexer`: add `@opentelemetry/api@^1.9.0` as a direct dependency to satisfy the new peer requirement.
- `pnpm-lock.yaml`: regenerated to reflect the dependency reshuffling.


PedroBinotto (approved):
lgtm

apps/gateful/package.json
```diff
@@ -24,6 +24,7 @@
   "dependencies": {
     "@anticapture/observability": "workspace:*",
     "@hono/node-server": "^1.14.0",
+    "@opentelemetry/api": "^1.9.0",
```
pikonha, line 27: why is this dependency needed? is it used anywhere?

LeonardoVieira1630: Yes. The observability package needs this. Then, instead of importing over there, it is being defined as a peer-dependency. 

I did it like that because this makes it impossible to do a "duble import". If gateful, for ex, use the @anticapture/observability and also the opentelemetry/api lib, the package.json would have duplicated dependency. 

This is why peer-dependency exists

PR #1838, opened 2026-04-24
Create Proposal

brunod-e:


isadorable-png:
## UI Review

Figma-grounded review against `mUgy2KpQ3gJ07yZaUaXu8l` node `2576:92794`. Vercel preview is auth-protected, so live-preview confirmation wasn't possible — findings are diff + Figma.

---

### Must-fix

**1. `handleNewProposal` hardcodes the whitelabel path**
`GovernanceSection.tsx` does `router.push('/whitelabel/${daoId}/proposals/new')` literally, ignoring the `basePath` already in scope. Same root issue affects `NavigationGuard.allowedPathname`.
→ Use `${basePath}/proposals/new` in both spots.

**2. Progress bar / `canPublish` gate counts the wrong fields**
Required fields are **Title + Description + Actions**. Discussion URL is optional. Current code has `filledCount = title + discussionUrl + body`, which produces two visible bugs:
- Filling Discussion URL ticks the bar to 3/3 even when Description or Actions are empty.
- Adding Actions doesn't move the bar at all.

→ Change to `filledCount = title + body + (actions.length > 0)` and update `canPublish` to follow. Also tighten the schema: `actions: z.array(...).min(1, "At least one action is required")` so form-level `isValid` aligns with the gate.

**3. No mobile back-navigation on `/proposals/new`**
`WhitelabelHeader` is `hidden lg:flex`, so the breadcrumb never renders below 1024px. Figma mobile frame `2576:93315` shows a "Proposals / New Proposal" breadcrumb header. Mobile users have no breadcrumb, no back button, and no Cancel in the sticky bottom bar — only Save Draft and Publish.
→ Render the breadcrumb on mobile for this route, or add a back/cancel affordance inside `ProposalCreationForm`.

**4. Mobile Drawer missing rounded top corners**
The Modal switches to a vaul Drawer below `md`. Figma specs every mobile drawer frame with `topLeftRadius: 8, topRightRadius: 8`. Code's `DrawerPrimitive.Content` only has `border-t`.
→ Add `rounded-t-base` (or `rounded-t-[8px]`) to the drawer content class.

**5. `useScreenSize` SSR hydration flash**
`useScreenSize` initializes `isMobile: false` and only reads `window.matchMedia` inside `useEffect`. The server always renders a Dialog; mobile clients re-render as a Drawer on hydrate.
→ Options: (a) `isMounted` guard so neither renders until known, (b) CSS-only show/hide via `md:hidden`/`md:block`, or (c) `suppressHydrationWarning` + cookie hint.

---

### Nice-to-haves

- **`BodyField` toolbar `flex-col-reverse` on mobile** — flips formatting buttons above the Visual/Markdown tabs. Figma mobile shows tabs first, then formatting.
- **`BodyField` `min-h-[520px]` unconditional** — eats most of a phone viewport. Make responsive: `min-h-[300px] md:min-h-[520px]`.
- **`BODY_PLACEHOLDER` strict-equality check is fragile** — MDX editors often normalize whitespace, so `values.body !== BODY_PLACEHOLDER` can fail to catch unmodified placeholder content. Better: track an explicit `bodyTouched` flag.
- **`?tab=drafts` URL stale when disconnected** — drafts tab only renders when connected, but the URL state survives disconnect. Reset to `proposals` when `!isConnected`.
- **Duplicate `daoConfig` import in `GovernanceSection.tsx`** — same default export imported twice under different names. Collapse to one.
- **USD column width-shifts Amount field in `AddTransferModal`** — Amount is full-width when `showUsd` is false, then 50/50 when true. Switching ETH ↔ ERC-20 toggles this, jumping layout. Reserve the column or stack USD below.
- **ERC-20 token-address validation fires on every keystroke** — error appears at character 1 and persists until the full 42-char address is entered. Validate on `onBlur` or after a debounce.
- **Sticky bottom nav lacks iOS safe-area inset** — add `pb-[env(safe-area-inset-bottom)]` so buttons stay tappable on notched iPhones.
- **`Select` `side={"bottom"}` brace style** — should be `side="bottom"` (string literal, no JSX expression).

---

### Open questions

1. **`InsufficientVPModal` "...on ENS"** — both code and Figma hardcode "ENS". `usePublishProposal` early-returns for non-ENS DAOs in v1, so this is currently scoped to ENS only. When the feature unlocks for other DAOs, will the copy be parameterized via `daoConfig.name`? Worth deciding now.

2. **Mobile back-nav implementation** — render the breadcrumb on mobile inside `WhitelabelHeader`, or add a back button inside `ProposalCreationForm`?

3. **`useScreenSize` SSR strategy** — does the team have a standard pattern for responsive component mounting elsewhere in the codebase? Whatever you pick here sets precedent.

---

The following looked off in code but match the design intent:
- PublishModal "Confirm your vote in your wallet" / "Wait for vote submission" wording
- `AddTransferModal` USD field marked required
- `StepList` wrapper without `rounded-base` (Figma container has `cornerRadius: 0`)
- `StepRow` inner border ring (intentional `cornerRadius: 100` "Border" node)
- Mobile Drawer without drag handle
- `DraftCard` Edit + Delete both outline variant
- ABI dropzone only after fetch error

isadorable-png (changes requested):
UI Improvements:
- Add a soft slide-in animation when clicking on "create proposal", so it doesn't feel too abrupt the page change.
- on the inline banner alert: " A minimum of 100000 is required." use "100K" and the DAO token (in this case 100K ENS).
- The input component that we have on the proposal body and in the proposal modal are different, let's use the same style for both (the one in the modal looks better since is white and it does not look like it's disabled).
- Add a margin bottom to the text editor header.
Image
- The description input should have a max-height, right now if I keep on writing it will get bigger and bigger. The textarea should have a inner scroll.
- Do not allow the user to type inside the ABI textearea unless there's an error fetching the ABI.
Image
- Switch on the mobile version the order: put the tabs first (that should have full-width) and then the edition options. Otherwise, when toggling, it keeps going up and down.
Image

apps/dashboard/features/create-proposal/hooks/useProposalThreshold.ts
```diff
@@ -0,0 +1,34 @@
...
+import daoConfig from "@/shared/dao-config";
+import type { DaoIdEnum } from "@/shared/types/daos";
+
+export type UseProposalThresholdReturn = {
+  threshold: bigint | null;
+  thresholdFormatted: string | null;
+  isLoading: boolean;
+};
+
+export function useProposalThreshold(
```
pikonha: use the `@anticapture/client` instead

apps/dashboard/features/create-proposal/hooks/useProposalVotingPower.ts
```diff
@@ -0,0 +1,39 @@
...
+  daoId: string,
+  address: string | undefined,
+): UseProposalVotingPowerReturn {
+  const daoIdEnum = daoId.toUpperCase() as DaoIdEnum;
+  const { data, loading: isVpLoading } = useGetVotingPowerQuery({
+    variables: { address: address ?? "" },
+    context: { headers: { "anticapture-dao-id": daoIdEnum } },
+    skip: !address,
+  });
+  const { threshold, isLoading: isThresholdLoading } =
```
pikonha: the voting power should be at the time of the proposal snapshot.

apps/dashboard/features/create-proposal/hooks/usePublishProposal.ts
```diff
@@ -0,0 +1,145 @@
...
+import { parseEventLogs, type Abi } from "viem";
+import {
+  usePublicClient,
+  useWaitForTransactionReceipt,
+  useWriteContract,
+} from "wagmi";
+import { mainnet } from "wagmi/chains";
+// TODO(multi-dao): load governor ABI from
+// daoConfig[daoIdEnum].daoOverview.contracts.governor once we support non-ENS
+// DAOs. Today this hook only supports ENS — see the assertion in `publish`.
```
pikonha: does any of the current daos that we provide the governance frontend have different create proposal signatures?

apps/dashboard/.env.example
```diff
@@ -22,3 +22,7 @@ GITHUB_TOKEN=
 
 # Blockful API token (server-only, injected by proxy)
 BLOCKFUL_API_TOKEN=
+
+# Etherscan v2 API key (used to fetch contract ABIs for custom proposal actions)
+# https://etherscan.io/apis
+NEXT_PUBLIC_ETHERSCAN_API_KEY=
```
pikonha: this will leak the etherscan api key, it would be better to move it to the next proxy the same way it is done for the [gateful calls](https://github.com/blockful/anticapture/blob/dev/apps/dashboard/app/api/gateful/%5B...path%5D/route.ts)

apps/dashboard/features/create-proposal/types/index.ts
```diff
@@ -0,0 +1,43 @@
...
+export type ProposalDraft = {
+  id: string;
+  daoId: string;
+  createdAt: number;
+  updatedAt: number;
+  title: string;
+  discussionUrl: string;
+  body: string;
+  actions: ProposalAction[];
+};
```
pikonha: coldn't this be gathered from the `ProposalOnchain` imported from @anticapture/client? 

I want to avoid duplicated types as much as possible,

apps/dashboard/features/create-proposal/hooks/usePublishProposal.ts
```diff
@@ -0,0 +1,145 @@
...
+  const proposalId =
+    receipt && isReceiptSuccess
+      ? ((
+          parseEventLogs({
+            abi: governorAbi,
+            logs: receipt.logs,
+            eventName: "ProposalCreated",
+          })[0]?.args as { proposalId?: bigint } | undefined
+        )?.proposalId ?? null)
+      : null;
```
pikonha, line 155: what in the name of god is this?

apps/dashboard/features/create-proposal/hooks/useProposalVotingPower.ts
```diff
@@ -0,0 +1,72 @@
...
+    args:
+      address && snapshotBlock !== undefined
+        ? [address as `0x${string}`, snapshotBlock]
+        : undefined,
+    query: {
+      enabled: Boolean(
+        address && governorAddress && snapshotBlock !== undefined,
+      ),
+    },
+  });
```
pikonha: this should be done through the API since we already have this data indexed

apps/dashboard/features/create-proposal/hooks/useProposalVotingPower.ts
```diff
@@ -0,0 +1,72 @@
...
+  return useMemo(() => {
+    const isLoading = isVpLoading || isThresholdLoading;
+    const isKnown = !isLoading && votesRaw !== undefined && threshold != null;
+    return {
+      votingPower: votesRaw ?? 0n,
+      threshold,
+      hasEnough: isKnown ? votesRaw >= threshold : null,
+      isLoading,
+    };
+  }, [votesRaw, threshold, isVpLoading, isThresholdLoading]);
```
pikonha: those memoized values only add complexity with literally no gain, return the plain object

apps/dashboard/features/create-proposal/hooks/useProposalVotingPower.ts
```diff
@@ -0,0 +1,72 @@
...
+    type: "function",
+    name: "getVotes",
+    stateMutability: "view",
+    inputs: [
+      { name: "account", type: "address" },
+      { name: "blockNumber", type: "uint256" },
+    ],
+    outputs: [{ name: "", type: "uint256" }],
+  },
+] as const satisfies Abi;
```
pikonha:
```suggestion
```

apps/dashboard/features/create-proposal/hooks/useProposalVotingPower.ts
```diff
@@ -0,0 +1,72 @@
...
+import { useProposalThreshold } from "@/features/create-proposal/hooks/useProposalThreshold";
+import daoConfig from "@/shared/dao-config";
+import type { DaoIdEnum } from "@/shared/types/daos";
+
+export type UseProposalVotingPowerReturn = {
+  votingPower: bigint;
+  threshold: bigint | null;
+  hasEnough: boolean | null;
+  isLoading: boolean;
+};
```
pikonha:
```suggestion
```

apps/dashboard/features/create-proposal/hooks/useProposalThreshold.ts
```diff
@@ -0,0 +1,50 @@
...
+    const raw = data?.proposalThreshold;
+    if (!daoPathParam || raw == null) {
+      return { threshold: null, thresholdFormatted: null, isLoading };
+    }
+    const threshold = BigInt(raw);
+    const daoIdEnum = daoId.toUpperCase() as DaoIdEnum;
+    const decimals = daoConfig[daoIdEnum]?.decimals ?? 18;
+    const formatted = formatUnits(threshold, decimals);
+    return { threshold, thresholdFormatted: formatted, isLoading };
+  }, [data, daoPathParam, daoId, isLoading]);
```
pikonha: return the plain object

apps/dashboard/features/create-proposal/utils/encodeActions.ts
pikonha, line 1: what does this file even do?

apps/dashboard/features/create-proposal/constants.ts
```diff
@@ -0,0 +1,32 @@
+import { DaoIdEnum } from "@/shared/types/daos";
+
+export const BODY_CHAR_LIMIT = 10_000;
+export const BODY_WARNING_THRESHOLD = 9_500;
+
+export const canCreateProposalForDao = (daoId: DaoIdEnum | null | undefined) =>
+  daoId === DaoIdEnum.ENS;
```
pikonha, line 7: shouldn't this be a dao config property?

apps/dashboard/features/create-proposal/schema.ts
```diff
@@ -0,0 +1,94 @@
...
+  .min(1, "Required")
+  .refine((v) => {
+    const trimmed = v.trim();
+    return isAddress(trimmed) || isEnsAddress(trimmed);
+  }, "Must be a valid address or ENS name");
+
+const strictAddressSchema = z
+  .string()
+  .min(1, "Required")
+  .refine((v) => isAddress(v.trim()), "Must be a valid Ethereum address");
```
pikonha, line 17: i guess this could be case insensitive

```suggestion
  .refine((v) => isAddress(v.trim(). {strict: false}), "Must be a valid Ethereum address");
```

apps/dashboard/app/whitelabel/[daoId]/proposals/new/page.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+export default async function NewProposalPage({ params }: Props) {
+  const { daoId } = await params;
+  const daoIdEnum = toDaoIdEnum(daoId);
+
+  if (!canCreateProposalForDao(daoIdEnum)) {
+    notFound();
+  }
+
+  return <ProposalCreationForm isWhitelabelRoute />;
```
Comment (pikonha) [inline apps/dashboard/app/whitelabel/[daoId]/proposals/new/page.tsx:42]: couldn't the dao id from param be parsed in here and passed through props instead of parsed directly on the [hook](https://github.com/blockful/anticapture/pull/1838/changes#diff-7036c4802dc630359a3309992e4306816e82b812f27e275dccc387eb5ed20ff5R35)?

pikonha (approved):

PR #1840, opened 2026-04-27
Gateful 🤝 address enrichment

pikonha:
## Summary
- merge the address-enrichment OpenAPI spec into Gateful's `/docs/json` output under the `/address-enrichment` prefix
- include address-enrichment schemas in the merged gateway spec and keep the existing spec intact when that upstream doc is unavailable
- add regression tests covering both the successful merge path and the upstream-failure fallback
- update `scripts/dev.sh` to call the `railway` CLI directly instead of `pnpm railway`

### DIFF
--- a/apps/gateful/openapi/gateful.json
+++ b/apps/gateful/openapi/gateful.json
@@ -1986,7 +1986,37 @@
       "get": {
         "responses": {
           "200": {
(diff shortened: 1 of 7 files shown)


PedroBinotto (approved):
lgtm

PR #1855, opened 2026-05-06
Feat/relayer public config endpoints

LeonardoVieira1630:
## Summary

Adds two read-only endpoints to the relayer (`GET /config` and `GET /rate-limit/:address`) so the dashboard can render eligibility hints and per-user remaining quota without hardcoding values or consuming the rate-limit budget on every check. Extends `RateLimitStorage` with a non-consumptive `getCount` method to back the rate-limit endpoint.

## Changes

- Added `GET /config` returning `minVotingPower` (decimal string preserving uint256 precision) and `maxRelayPerAddressPerDay`.
- Added `GET /rate-limit/:address` returning per-operation `used`/`remaining` counts for `vote` and `delegation`, plus `resetsAt` (next UTC-midnight ISO timestamp).
- Extended `RateLimitStorage` interface with `getCount` and `RedisClient` with `get`; implemented `RedisRateLimitStorage.getCount` using a plain `GET` so reads do not increment the counter.
- Added new schemas (`schemas/config.ts`, `schemas/rate-limit.ts`) following the existing `@hono/zod-openapi` pattern, with `z.iso.datetime()` for `resetsAt` to surface `format: date-time` in the OpenAPI spec.
- Wired both routes in `src/index.ts` so they appear in `/docs`.
- Updated existing storage mocks (`rate-limiter.test.ts`, `e2e/helpers/app.ts`) to satisfy the extended interface.
- Added 3 storage tests covering: empty key returns 0, repeated reads do not consume the limit, and entries from previous UTC days are ignored.

### DIFF
--- a/apps/relayer/src/schemas/config.ts
+++ b/apps/relayer/src/schemas/config.ts
@@ -2,17 +2,15 @@ import { z } from "@hono/zod-openapi";
 
 export const ConfigResponseSchema = z
   .object({
-    minVotingPower: z
-      .string()
-      .describe(
+    minVotingPower: z.string().openapi({
+      format: "bigint",
+      description:
         "Minimum voting power required to relay, as a decimal string (uint256).",
-      ),
-    maxRelayPerAddressPerDay: z
-      .number()
-      .int()
-      .min(0)
-      .describe(
+    }),
+    maxRelayPerAddressPerDay: z.number().int().min(0).openapi({
+      format: "bigint",
+      description:
         "Maximum number of relays per address per UTC day, per operation.",
-      ),
+    }),
   })
   .openapi("RelayerConfigResponse");
--- a/apps/relayer/src/schemas/config.ts
+++ b/apps/relayer/src/schemas/config.ts
@@ -1,0 +1,18 @@
+import { z } from "@hono/zod-openapi";
+
+export const ConfigResponseSchema = z
+  .object({
+    minVotingPower: z
+      .string()
+      .describe(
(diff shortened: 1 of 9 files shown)


apps/relayer/src/schemas/config.ts
```diff
@@ -0,0 +1,18 @@
...
+      .describe(
+        "Minimum voting power required to relay, as a decimal string (uint256).",
+      ),
+    maxRelayPerAddressPerDay: z
+      .number()
+      .int()
+      .min(0)
+      .describe(
+        "Maximum number of relays per address per UTC day, per operation.",
+      ),
```
pikonha:
```suggestion
      .openapi({format:"bigint", description: "Maximum number of relays per address per UTC day, per operation."}),
```

apps/relayer/src/schemas/config.ts
```diff
@@ -0,0 +1,18 @@
+import { z } from "@hono/zod-openapi";
+
+export const ConfigResponseSchema = z
+  .object({
+    minVotingPower: z
+      .string()
+      .describe(
+        "Minimum voting power required to relay, as a decimal string (uint256).",
+      ),
```
pikonha: this will lead to kubb generating it as bigint instead of string (as soon as the ethereum generator gets merged :p)

```suggestion
     .openapi({format:"bigint", description: "Minimum voting power required to relay, as a decimal string (uint256)."}),
```

pikonha (approved):

brunod-e (approved):

PR #1875, opened 2026-05-10
chore: migrate event metadata

PedroBinotto:


isadorable-png:
## 🎨 UI Review

> Automated review · [Spec: DEV-555](https://app.clickup.com/t/86agc8kwf) · No Figma reference found · Preview: auth-gated (401)
> ℹ️ Reviewed without a Figma reference — grounded in DEV-555 spec text + UX-expert evaluation. Vercel preview returned 401 (SAML-protected, no bypass token); regression check limited to diff-level analysis.

---

### Activity Feed Filters — Event Type section

**Default filter state doesn't match spec** ← must-fix
- DEV-555 says: *"Default state: all event types checked, except transfers"*
- Current code: `parseAsArrayOf(...)` has no `defaultValue`, so on first load `type` is `undefined` → the "All" radio is selected → **transfers are shown by default**
- Fix: provide a default in `useActivityFeedParams.ts:36` so the initial state excludes `TRANSFER`:
  ```ts
  parseAsArrayOf(
    parseAsStringEnum<FeedEventType>(Object.values(feedEventTypeEnum)),
  ).withDefault([
    feedEventTypeEnum.VOTE,
    feedEventTypeEnum.PROPOSAL,
    feedEventTypeEnum.PROPOSAL_EXTENDED,
    feedEventTypeEnum.DELEGATION,
  ])
  ```
  Also initialise `localFilters` in the drawer's `useEffect` to reflect that default on open. `[Code-only]`

---

**Checkbox label text is always `text-primary` — doesn't track checked state** ← must-fix
- `<span className="text-primary text-sm">` is hardcoded regardless of whether the checkbox is checked (`ActivityFeedFilters.tsx:221`)
- `RadioButton` uses `text-secondary` (unchecked) → `text-primary` (checked/hover). In the same vertical list the "All" label dims when unchecked while the checkbox labels stay full-brightness — the two rows look visually inconsistent
- Fix: make the label color state-aware:
  ```tsx
  <span className={cn("text-sm transition-colors duration-200", checked ? "text-primary" : "text-secondary")}>
    {label}
  </span>
  ```
  (`ActivityFeedFilters.tsx:221`) `[Code-only]`

---

**Missing hover transition on Checkbox label** ← nice-to-have
- The `<label>` wrapper has no `group` class; the `<span>` has no `group-hover:text-primary` — hovering the row produces no label-text feedback, unlike `RadioButton` which brightens on hover
- Fix: add `group` to the `<label>` and `group-hover:text-primary` to the `<span>` (`ActivityFeedFilters.tsx:215`) `[Code-only]`

---

**Mixed RadioButton + Checkbox visual pattern in the same section** ← question for author
- "All" = circular radio indicator; the five type options = square checkbox indicators — two different control shapes in a single vertical list
- This is a non-standard pattern that could signal "pick one *or* pick many" ambiguously. A common alternative is to keep "All" as a checkbox too (checking it selects all, unchecking clears all)
- Not blocking, but worth a second look if testing shows users are confused `[UX-heuristic]`

---

### Validated against spec (DEV-555) — no change needed

- Using `Checkbox` for individual type options is correct per spec ("must be checkboxes — users need to select multiple event types simultaneously")
- `handleTypeToggle` logic is sound: adds/removes types and collapses to `undefined` when the last one is unchecked
- `getActiveActivityFeedFiltersCount` correctly counts a non-empty `type[]` as one active filter (not N filters)
- Removing `DELEGATION_VOTES_CHANGED` from the enum and feed event type selector is intentional per spec


---

apps/api/sqlcheck.ts
pikonha, line 1: what does this file do?

apps/api/src/mappers/feed/index.ts
pikonha: this could be a [anyOf](https://swagger.io/docs/specification/v3_0/data-models/oneof-anyof-allof-not/) openapi property so it is correctly parsed by kubb generating a valid TS type

apps/api/src/repositories/feed/index.ts
pikonha, line 1: it is already a pretty expensive database query. How do these changes affect it?

apps/api/src/services/feed/index.ts
```diff
@@ -3,12 +3,16 @@ import { DaoIdEnum } from "@/lib/enums";
 import { getDaoRelevanceThreshold } from "@/lib/eventRelevance";
 import { DBFeedEvent, FeedRequest, FeedResponse } from "@/mappers";
 
+type FeedEventWithMetadata = DBFeedEvent & {
+  metadata: Record<string, unknown> | null;
+};
```
pikonha, line 15: these is both duplicated and not handling the anyOf spec, so it would lead to a unknown object on the TS codegen

pikonha (changes requested):

apps/api/sqlcheck.ts
PedroBinotto: had claude write it up to test some DB changes; removed!

apps/api/src/mappers/feed/index.ts
PedroBinotto: It looks a bit cumbersome but for practical reasons it probably works better than the previous "anonymous" object type

pikonha (approved):

isadorable-png (approved):

PR #1888, opened 2026-05-12
Feat: MCP Enhancements

PedroBinotto:
- Depends on https://github.com/blockful/notification-system/pull/258

---

## Summary

Refines the MCP server and supporting REST surface so LLM clients get smaller, better-described payloads.

- **Lean proposal endpoints** — adds `GET /proposals/lean`, `/proposals/lean/{id}`, `/proposals/lean/search` (and Snapshot equivalents under `/offchain/proposals/lean*`). These mirror the existing endpoints but omit heavy
fields (`calldatas`/`values`/`targets` on-chain, `body` off-chain). The MCP server (`packages/anticapture-client/src/create-mcp-server.ts`) now binds the `proposals`, `searchProposals`, `proposal`, and Snapshot equivalents to
the lean variants. The full-payload endpoints remain available over REST.
- **Structured health** — `GET /health` now returns a snapshot (`status`, `database`, `chain.head`, `indexer.{lastEventTimestamp,lagSeconds,fresh}`). Freshness threshold is 300s. New gateway route `GET /{dao}/health` proxies
the per-DAO snapshot and includes the gateway's circuit-breaker state.
- **`rawDelta`** added to every change-rate response (governance-activity, token-distribution) so clients don't need to recompute the signed delta themselves. `changeRate` precision bumped from 2 to 6 decimals.
- **OpenAPI polish** — operationIds/summary/tags added to gateway-owned routes (`gatewayHealth`, `daos`, `averageDelegationPercentage`); DAO identifier descriptions clarified (`uppercase` for response IDs, `case-insensitive`
for path params); merged upstream `/health` paths blocklisted so the gateway's resolver definition wins.

## ⚠️ Breaking changes

These rename/shape changes will require coordinated client updates:

1. **`DaoResponse` field rename** — `alreadySupportCalldataReview` → `supportsCalldataReview`, `supportOffchainData` → `supportsOffchainData`. Affects `/{dao}/dao`, `/daos`, the GraphQL schema, the MCP `dao`/`daos` tools, and
any generated client types. Dashboard and other consumers must update field references.
2. **Voting-power path parameter rename** — `/voting-powers/{accountId}` → `/voting-powers/{address}`. The URL itself still matches positionally, but OpenAPI-generated clients and the MCP `votingPowerByAccountId` tool now
expect `address` in place of `accountId`.
3. **`/health` 503 response shape** — the legacy `{ status, database, message }` body is gone. The error response is now the full snapshot shape with `database: "error"`. Consumers that surfaced the `message` string lose that
text.


isadorable-png:
## 🎨 UI Review

> Automated review · No Figma reference found · Preview: Vercel build failed (Error on both deployments)
> ℹ️ **Path A** — Reviewed without a Figma reference. Grounded in diff analysis + UX-expert evaluation.
> ⚠️ **Vercel preview unavailable** — both `anticapture` and `anticapture-storybook` deployments are in Error state on this PR.

---

### Dashboard files — no visual changes

Both dashboard files in this diff are pure code formatting rewraps with zero visual or behavioral consequences:

- `DaoOverviewSection.tsx` — multi-line import block reformat only. No logic changed. ✓
- `OffchainVotingModal.tsx` — template literal line-wrap in `weightedValue` computation. No logic changed. ✓

---

### ⚠️ Breaking field rename not yet mirrored in dashboard — question for author `[Code-only]`

The PR renames two fields in `DaoResponse`:

| Old name | New name |
|---|---|
| `alreadySupportCalldataReview` | `supportsCalldataReview` |
| `supportOffchainData` | `supportsOffchainData` |

The GraphQL schema (`api-gateway/schema.graphql`) is updated in this PR, but the dashboard's GraphQL queries and generated client types have not been updated — no codegen run is included in the diff. If the dashboard queries these fields by their old names, they will silently return `undefined` after deployment, causing any gating logic that reads them (e.g. show/hide the calldata review tab, show/hide offchain data sections) to fail closed — those features would go missing for all DAOs without an error or console warning.

**Question for author:** Is the dashboard field-reference update in a separate follow-up PR? If so, these two PRs need to be merged and deployed together to avoid a regression window where the dashboard is broken. Please link the companion PR here so reviewers and deployers know to coordinate.

---

### Regression check

Vercel builds for both `anticapture` and `anticapture-storybook` are in **Error** state on this PR — no live preview comparison was possible. A clean build before merge is a prerequisite; the build failure likely relates to the breaking API shape changes introduced here.

---

*Reviewed by @isadorable-png — automated UI review*

---

apps/api/src/services/health/index.ts
```diff
@@ -0,0 +1,55 @@
...
+  };
+}
+
+export class HealthService {
+  constructor(
+    private readonly repo: HealthRepository,
+    private readonly daoClient: DAOClient,
+  ) {}
+
+  async getHealth(): Promise<HealthReport> {
```
pikonha, line 34: I'm not sure [Railway's healthcheck](https://docs.railway.com/deployments/healthchecks) can handle anything other than an HTTP status code to know whether the API is working or not. I thus recommend you to handle errors using an HTTP code instead of a property

pikonha (changes requested):

apps/gateful/src/health/dao.ts
pikonha, line 1: what's this used for? it has the same Railway healthcheck problem as the API

apps/gateful/openapi/gateful.json
```diff
@@ -353,17 +453,25 @@
         "properties": {
           "previousValue": {
             "type": "string",
-            "format": "bigint"
```
pikonha, line 356: do not remove it

packages/anticapture-client/mcp-server.ts
pikonha, line 1: What's changed in this file's behavior?

apps/api/src/controllers/account-balance/listing.ts
```diff
@@ -125,6 +134,13 @@ export function accountBalances(
         toTimestamp,
       );
 
+      if (!result) {
+        return context.json(
+          ErrorResponseSchema.parse({ error: "Account not found" }),
+          404,
+        );
+      }
```
pikonha, line 142: ensure this doesn't break the old gateway codegen

apps/api/src/controllers/proposals/offchainProposals.ts
```diff
@@ -92,6 +95,150 @@ export function offchainProposals(
...
+        200,
+      );
+    },
+  );
+
+  app.openapi(
+    createRoute({
+      method: "get",
+      operationId: "offchainSearchProposalsLean",
+      path: "/offchain/proposals/lean/search",
```
pikonha: couldn't this be a query parameter instead of a new endpoint?

apps/address-enrichment/src/controllers/mappers.ts
```diff
@@ -78,10 +78,19 @@ export const AddressResponseSchema = z.object({
...
+        return Array.isArray(val) ? val : [val];
+      },
+      z
+        .array(AddressSchema)
+        .min(1, "At least one address is required")
+        .max(100, "Maximum 100 addresses per request"),
+    )
     .openapi({
+      type: "array",
+      items: { type: "string" },
```
pikonha:
```suggestion
      items: { type: "string", format: "ethereum-address" },
```

apps/api/src/controllers/governance-activity/controller.ts
```diff
@@ -95,21 +95,23 @@ export function governanceActivity(
...
       }
-      const changeRate =
-        data.oldProposalsLaunched &&
-        data.currentProposalsLaunched / data.oldProposalsLaunched - 1;
+      const changeRate = data.oldProposalsLaunched
+        ? Number(
+            (
+              data.currentProposalsLaunched / data.oldProposalsLaunched -
+              1
+            ).toFixed(6),
```
pikonha, line 108: do we want a 6 decimals value?

apps/api/src/controllers/governance-activity/controller.ts
```diff
@@ -145,20 +147,18 @@ export function governanceActivity(
...
           },
           200,
         );
       }
 
-      const changeRate = data.oldVotes && data.currentVotes / data.oldVotes - 1;
+      const changeRate = data.oldVotes
+        ? Number((data.currentVotes / data.oldVotes - 1).toFixed(6))
+        : 0;
+      const rawDelta = data.currentVotes - data.oldVotes;
```
pikonha, line 159: why is this rawDelta used for? and what's the difference from the value we already had?

pikonha (changes requested):

apps/gateful/src/health/dao.ts
PedroBinotto: This was changed for the sake of MCP dao healthchecks and service discovery; I'll just split the concerns into two different endpoints so that no regression is made

packages/anticapture-client/mcp-server.ts
PedroBinotto: `stdio` clients (people running the MCP server locally instead of connecting to our public URL) are to connect to the lean variants of the proposal endpoints + some adjustments to allow access to cross-DAO aggregation tools (avgDelegation etc.)

apps/api/src/controllers/governance-activity/controller.ts
PedroBinotto: > why is this rawDelta used for? 

The MCP may use it; 

> and what's the difference from the value we already had?

This is just the raw difference in wei, not percentage points

pikonha (approved):

isadorable-png (approved):

PR #1896, opened 2026-05-12
86af3519p: perf: fix derived state in effects and remove isMounted pattern

pikonha:
## Summary

Addresses unnecessary re-render cycles caused by derived state computed inside `useEffect` hooks, a redundant `isLoaded`-driven sync effect, a missing lazy state initializer, an uncleaned debounce timeout, and a `useMemo` wrapper around a trivial arithmetic expression.

ClickUp task: https://app.clickup.com/t/86af3519p (DEV-401)

---

## Approach

Each fix was kept minimal and surgical — no logic was restructured beyond what was required to eliminate the anti-pattern.

---

## Files changed and why

| File | Change |
|---|---|
| `apps/dashboard/shared/hooks/useMessageStack.ts` | Added `isLoadedRef` (`useRef`) as a guard in the localStorage sync effect. Previously the sync effect depended on `isLoaded` state, so it would fire a second time when `isLoaded` transitioned to `true` — even if `closedMessageIds` hadn't changed (empty-storage case). The ref is set synchronously in the load effect; `isLoaded` state is kept for the hook's return value. |
| `apps/dashboard/features/holders-and-delegates/hooks/useProposalsActivity.ts` | Removed `accumulatedProposals` state and the `useEffect` that set it. `fetchMore`'s `updateQuery` already accumulates proposals in the Apollo cache, so `data.proposalsActivity.proposals` already holds all pages. `processedData` now derives proposals directly from `data`; `fetchNextPage` uses `data?.proposalsActivity?.proposals?.length` for the `skip` offset. |
| `apps/dashboard/features/holders-and-delegates/hooks/useDelegates.ts` | Changed `useState(new Map())` and `useState(new Set())` to `useState(() => new Map())` / `useState(() => new Set())`. Without a lazy initializer, React allocates a new `Map`/`Set` on every render (only the first value is used, but the allocation still happens). |
| `apps/dashboard/features/token-distribution/components/SectionComposedChart.tsx` | Added a cleanup `useEffect` that calls `clearTimeout(brushTimeoutRef.current)` on unmount, preventing a pending 100 ms debounce from firing against a dead component. |
| `apps/dashboard/features/holders-and-delegates/hooks/useDelegateDelegationHistory.ts` | Inlined `hasNextPage` as a plain expression — `currentPage * limit < totalCount`. The previous `useMemo` wrapper added overhead (dependency tracking, closure allocation) with no benefit for a single arithmetic comparison. |

---

## Assumptions I made

- Apollo's `updateQuery` in `fetchNextPage` is the source of truth for accumulated proposals; the separate `accumulatedProposals` state was a redundant cache of that data.
- `isLoaded` must remain as React state (not converted to a ref) because it is part of the hook's public API and consumers may render based on it.
- The `SectionComposedChart` component is purely presentational; the cleanup effect has no visible behavior change.

---

## What I did NOT do

- **`useTokenHolders.ts:74-82`** — No `useEffect` calls exist in the current version of this file; the described "two separate effects resetting cache" appear to have already been removed. No change made.
- **`VotingPowerTable.tsx`** — File not found anywhere in the repository; the `isMounted` anti-pattern described in the task may have already been addressed. No change made.
- **`useVotingPower.ts`** — File not found anywhere in the repository. No change made.
- Did not add, skip, or remove any tests.
- Did not touch `.github/`, `infra/`, or `.env*` files.

---

## Open questions for review

1. **`useProposalsActivity`** — Is there any scenario where the Apollo cache does _not_ accumulate proposals correctly via `updateQuery`, making the removed `accumulatedProposals` state necessary? If so, the state should be restored.
2. **`useMessageStack` / SSR** — The load effect still runs on the client only (correct, since `localStorage` is client-side), and `isLoadedRef.current` is set synchronously before the re-render. Should `isLoaded` be kept in the return type, or can it be dropped in a follow-up to further simplify consumers?
3. **`useTokenHolders` / `VotingPowerTable` / `useVotingPower`** — Are these files genuinely gone (merged/renamed), or should they be looked up under different paths?

---
_Generated by [Claude Code](https://claude.ai/code/session_017LpXmFrE3aXSM9hhrqUb3D)_


pikonha:
**Re: `useMessageStack.ts` — reverted to original pattern (commit 883ab28)**

The `chatgpt-codex-connector` bot correctly identified a bug in my ref-based guard. Here's the analysis:

**The bug:** `isLoadedRef.current = true` was set synchronously inside the load effect body. Since React runs effects in declaration order within the same flush, the sync effect (declared second) would execute right after and find `isLoadedRef.current === true` — then immediately write the empty initial `closedMessageIds` to localStorage, overwriting any persisted data.

Under React StrictMode (default in Next.js dev mode), the mount→unmount→remount cycle would then read back that now-empty value on the second mount, permanently clearing the user's closed message IDs.

**Why the original code is correct:** The `if (isLoaded)` check in the sync effect uses the *closure-captured* `isLoaded = false` from the initial render. So on first mount, both effects run in the same flush, but the sync effect sees stale `isLoaded = false` and correctly skips. On the next render (after the batched state updates from the load effect), `isLoaded = true` and the correct `closedMessageIds` are both committed — the sync effect then writes valid data.

The "unnecessary write" the task described (writing `[]` when localStorage is empty) is cosmetic and not a correctness concern. The original code has been restored.

---

isadorable-png:
## 🎨 UI Review

> Automated review · [Spec: DEV-401](https://app.clickup.com/t/86af3519p) · Preview: auth-gated (403)
> ⚠️ Reviewed without a Figma reference — no matching design found in PR body or ClickUp (searched PR body → task custom fields → lists `901320776629` + `901323162483` → workspace).
> ℹ️ This PR contains no visual/UI changes — all diffs are pure hook logic files.

---

### No UI issues found

All four changed files are logic-only with zero rendered output changes:

| File | Change | Verdict |
|---|---|---|
| `useDelegateDelegationHistory.ts` | Inlined `hasNextPage` boolean arithmetic; removed `useMemo` wrapper | ✅ Correct — no DS overhead for a single comparison |
| `useDelegates.ts` | `useState(new Map())` → `useState(() => new Map())` (×2) | ✅ Correct — lazy initializer avoids per-render allocation |
| `useProposalsActivity.ts` | Removed `accumulatedProposals` state + effect; read directly from Apollo cache | ✅ Correct — see note below |
| `SectionComposedChart.tsx` | Added unmount cleanup `clearTimeout(brushTimeoutRef.current)` | ✅ Correct — prevents post-unmount state update |

No visual regressions are possible from any of these changes.

---

### Code notes (non-blocking)

**`useProposalsActivity` — Apollo accumulation assumption is sound**
The removal of `accumulatedProposals` state relies on `updateQuery` writing the fully-merged list back to Apollo cache. Reading the `fetchNextPage` callback confirms this: `[...prevItems, ...deduped newItems]` is returned as `proposalsActivity.proposals`, so `data?.proposalsActivity?.proposals` holds the full accumulated list after each `fetchMore`. `proposalsLength` therefore gives the correct `skip` offset on each subsequent call. The PR author's open question (comment #3) is a reasonable sanity check, but the code path is sound. [Code-only]

**`useDelegateDelegationHistory` — `|| 0` → `?? 0`**
The nullish coalescing change is a semantic improvement: `|| 0` would coerce `totalCount = 0` to `false` and substitute `0` anyway (functionally identical for this field), while `?? 0` is more precise. No behavior change in practice. [Code-only]

---

PR #1904, opened 2026-05-13
fix: create proposal fixes

brunod-e:
https://app.clickup.com/t/90132341641/DEV-659

### DIFF
--- a/apps/dashboard/features/create-proposal/components/modals/AddCustomActionModal.tsx
+++ b/apps/dashboard/features/create-proposal/components/modals/AddCustomActionModal.tsx
@@ -33,6 +33,123 @@ import ensGovernorAbi from "@/abis/ens-governor.json";
 type Step = 1 | 2;
 type ConfigMode = "fetch" | "calldata";
 
+type AbiTypeCategory =
+  | "bool"
+  | "uint"
+  | "int"
+  | "address"
+  | "bytes_dynamic"
+  | "bytes_fixed"
+  | "array"
+  | "string"
+  | "other";
+
+function getAbiTypeCategory(abiType: string): AbiTypeCategory {
+  if (abiType === "bool") return "bool";
+  if (/^uint\d*$/.test(abiType)) return "uint";
+  if (/^int\d*$/.test(abiType)) return "int";
+  if (abiType === "address") return "address";
+  if (abiType === "bytes") return "bytes_dynamic";
+  if (/^bytes\d+$/.test(abiType)) return "bytes_fixed";
+  if (abiType.endsWith("[]") || /\[\d+\]$/.test(abiType)) return "array";
+  if (abiType === "string") return "string";
+  return "other";
+}
+
+function getArgPlaceholder(abiType: string): string {
+  const cat = getAbiTypeCategory(abiType);
+  switch (cat) {
+    case "uint":
+      return "0 or 0x1a2b…";
+    case "int":
+      return "0 or -100";
+    case "address":
+      return "0x… or ENS name";
+    case "bytes_dynamic":
+      return "0x…";
+    case "bytes_fixed": {
+      const n = parseInt(abiType.replace("bytes", ""), 10);
+      return `0x… (${n} bytes)`;
+    }
+    case "array":
+      return '["item1", "item2"]';
+    case "string":
+      return "text…";
+    default:
+      return "";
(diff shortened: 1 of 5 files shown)


apps/dashboard/features/create-proposal/components/modals/AddCustomActionModal.tsx
pikonha, line 1: this seems rather complext, could you explain it?

isadorable-png:
## 🎨 UI Review

> Automated review · [Spec (ClickUp DEV-659)](https://app.clickup.com/t/86ahanmyq) · Preview (auth-gated 401, see caveats)
>
> ℹ️ **Path A** — Reviewed without a Figma reference (no Figma URL in PR body, DEV-659, or linked design task `86ahar8xb`). Grounded in 3 ClickUp task attachments + UX-expert evaluation.
> ⚠️ **Vercel preview returned 401** (SAML — no bypass token configured). Regression check and responsive verification skipped.

---

### Add Transfer Modal

**Spec 1.2 — Suggested tokens row missing** `[ClickUp-image-confirmed]`

The PR ships the spec 1.1 label change (`"ERC-20"` → `"ERC-20 Token"`) but does **not** add the suggested-tokens badge row described in spec 1.2. The ClickUp mockup shows `USDT • USDC • DAI • WETH • ENS` chips below the `Token Contract Address` field; selecting one should fill the input. The `AddTransferModal.tsx` diff is a single line — no badge row added.

→ Question for author: is this scoped to a follow-up PR? If yes, link the issue so this doesn't get re-flagged. If not, it's a missing acceptance criterion (`AddTransferModal.tsx:264-271`).

---

### Add Custom Action Modal — ABI fetch step

**ENS name replaced with hex address after lookup** `[UX-heuristic]`

When the user enters `walt.ens.eth`, `setContractAddress(resolved)` (`AddCustomActionModal.tsx:344-346`) overwrites the input with the 0x address — the typed ENS disappears from the visible input. Pro: clarifies which address was looked up. Con: loses readable ENS context (Etherscan-style UIs typically keep both visible).

→ Question for author: intentional? If yes, consider an inline helper like `Resolved: 0x… (walt.ens.eth)` so the ENS isn't lost. If unintentional, drop the auto-replace and pass `resolved` to `lookupAbi` directly.

---

### Help Popover

**Substring path match is too loose** `[Code-only]`

`pathname?.includes("/proposals/new")` (`HelpPopover.tsx:89`) matches any path *containing* that substring — a future `/proposals/newest` or `/proposals/new-thing` would also bump the help button up. Use `pathname === "/proposals/new"` for exact match, or `pathname?.startsWith("/proposals/new/")` if sub-routes exist.

**Position offset unverifiable without preview** `[UX-heuristic — preview-401]`

The fix uses `bottom-16` (64px) on the create-proposal route. Without an accessible preview I can't confirm the Publish CTA is fully reachable at that offset across viewports, especially on small screens where a floating button + bottom CTA bar can re-overlap. Worth a visual check before merge.

---

### Spec acceptance — delivered correctly (no change needed)

- ✅ **1.1** — `"ERC-20 Token"` label (`AddTransferModal.tsx:267`) matches the ClickUp mockup
- ✅ **2.1** — ABI fetch resolves ENS first, then uses the resolved address (`AddCustomActionModal.tsx:336-347`)
- ✅ **2.2** — function picker filters `view`/`pure`; status label says "N write function" (`AddCustomActionModal.tsx:291-294, 573`)
- ✅ **3.1** — `onWheel={e => e.stopPropagation()}` on `<Select>` lets the wheel scroll the listbox without scrolling the page (`Select.tsx:109`)
- ✅ **3.2** — Help popover shifts to `bottom-16` on `/proposals/new` (modulo the substring nit above)
- ✅ **4.1** — uint inputs gated to decimal or `0x` hex at keystroke; uint8/16/etc. validated against max
- ✅ **4.2** — bool args render as `<Select>` with `true`/`false` items
- ✅ **4.3** — arrays show `["item1", "item2"]` placeholder + JSON.parse validation
- ✅ **4.4** — address validates as hex or ENS; `bytesN` enforces hex + exact byte length
- ✅ Error display timing is good: `touchedArgs` Set defers errors until blur — no yelling while typing
- ✅ Selecting a new function resets both `args` and `touchedArgs` so stale state doesn't carry over

PR #1921, opened 2026-05-20
fix: improve revenue calculations and formatting in KpiRow and transform utilities

brunod-e:


isadorable-png (approved):
## UI Review — Path C (Figma MCP confirmed) `[Figma-confirmed]`

**Spec**: ClickUp `86af1vzrr` → Figma `mUgy2KpQ3gJ07yZaUaXu8l`
**Preview**: `anticapture-git-fix-revenue-issues-ful.vercel.app`
**Nodes checked**: KPI Row `2881:27089`, Monthly Revenue chart `2753:61715`, New Wallets chart `2881:27301`

---

### ✅ Changes confirmed against design

**`format.ts` — year label format** `[Figma-confirmed]`
Both charts (`2753:61715` and `2881:27301`) spec x-axis labels as 4-digit years: `2019 / 2020 / 2021 / 2022 / 2023 / 2024 / 2025`. The old `String(d.getUTCFullYear()).slice(2)` produced `"Dec 24"`, and the StackedBarChart formatter `20${value.split(" ")[1]}` then constructed `"202024"` for 4-digit inputs (broken). Both fixes together restore correct rendering.

**`StackedBarChart.tsx` — x-axis formatter** `[Code-only]`
`value.split(" ")[1] ?? value` correctly extracts `"2024"` from `"Dec 2024"`. Necessary internal change to stay consistent with `formatMonthLabel`.

**`transform.ts` — "New Wallets" trend computed** `[Figma-confirmed]`
Figma KPI Row (`2881:27089`) shows a directional arrow on the New Wallets card. Replacing the hardcoded `"up"` with `lastWallet.newWallets >= prevWallet.newWallets ? "up" : "down"` is correct. The `undefined` fallback when `prevWallet` is absent is also right (renders no arrow on first data point).

**`transform.ts` — division-by-zero guards** `[Code-only]`
`total > 0 ? ... : "0.0%"` across all three stream share calculations. No visual change under normal conditions; prevents `NaN` / `Infinity` renders on empty data (fresh deployment, loading states). ✅

**`KpiRow.tsx` — `monthStartUnix` helper** `[Code-only]`
Snaps rolling offset timestamps to UTC month boundaries before querying. Correct choice for a monthly-granularity chart — prevents partial-month windows from skewing period-over-period comparisons. One edge case to be aware of: the `"7d"` period label will snap back to the 1st of the previous month if today is early in the current month, so the visible window may be wider than 7 days.

---

### ⚠️ Pre-existing issues (not introduced by this PR — flag for follow-up)

**A — KPI up-arrow text uses a surface token instead of a text token** `[Figma-confirmed]`
Figma KPI Row (`2881:27089`) specifies `text-[color:var(--text/success,#4ade80)]` (bright green — `--base-success`) for improvement text. The current code uses `text-[#15803d]`, which is the `--surfaces/solid/success` **background** token. Result: up-arrows render as dark forest green rather than the spec's `#4ade80`.
→ Follow-up: replace `#15803d` with `var(--base-success, #4ade80)` on KPI arrow subtext.

**B — "Renewal Rate" subtext period mismatch** `[Figma-confirmed]`
Figma (`2881:27089`): `2pp vs. last quarter`
Code: `` `${renewalDelta.toFixed(0)}pp vs last month` ``
The spec's comparison window is a quarter; the code compares month-over-month.
→ Follow-up: align the copy with the actual calculation, or change the calculation to use quarterly delta.

---

**Verdict: Approve.** All four changes are correct. The Figma-relevant fixes (`formatMonthLabel`, StackedBarChart formatter, New Wallets trend direction) are verified against the design spec. Pre-existing issues A and B are out of scope for this PR.

---

pikonha (approved):

PR #1930, opened 2026-05-21
v2.2.5

pikonha:
> [!NOTE]
> **High Risk**
> Broad API and schema changes (health routing, feed metadata model, removed lean routes, pagination semantics) can break orchestrators and generated clients; feed synthesis and delegation window pagination are correctness-critical paths.
> 
> **Overview**
> **v2.2.5** bundles API contract and behavior fixes across `@anticapture/api`, gateway, gateful, indexer, and dashboard.
> 
> **Health:** `GET /health` is now a minimal liveness probe (`{ database }`, 200/503). Rich diagnostics (chain head, indexer lag, `status: ok|degraded|error`) move to **`GET /health/full`**. Gateful keeps public `/{dao}/health` but proxies **`/health/full`** upstream.
> 
> **Feed:** Drops stored `feed_event.metadata` in favor of **query-time synthesis** from on-chain tables; OpenAPI/GraphQL expose **discriminated metadata** types. `/feed/events` accepts **multi-value `type`** filters.
> 
> **Proposals:** Removes six dedicated `lean*` routes; existing list/search/by-id routes take **`lean=true`** to omit heavy fields (on-chain calldatas/values/targets, off-chain `body`), with explicit boolean parsing so `lean=false` stays false.
> 
> **Delegation percentage & aggregators:** DAO API drops cursor pagination and **`pageInfo`** on `/delegation-percentage`; uses **`skip`/`limit`** with **`totalCount`** over the full forward-filled window (no capped reads). Average-delegation aggregators switch from **`after`/`before`** to **`skip`** and derive **`hasNextPage`** from `items.length < totalCount`.
> 
> **Comparisons & types:** Supply/governance compare responses add **`rawDelta`** and tighter **`changeRate`** (6 decimals). **`percentageChange`** is a plain string (not OpenAPI bigint). DAO flags rename to **`supportsCalldataReview`** / **`supportsOffchainData`**. **`/voting-powers/{address}`** path param rename. Account balance by id returns **404** when missing. Address enrichment returns **502** on upstream failure.
> 
> **Dashboard (changeset):** Kubb SDK for governance; ENS revenue UI/metrics refinements.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 67d878df8d2bfde1ebd69ffdc614636692bc17c0. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


isadorable-png:
## 🎨 UI Review

> Automated review · Figma: not found — no link in PR body · Spec: not linked · Preview: `dev.anticapture.com` (SAML-protected, no bypass token)
> ℹ️ **Path A** — Reviewed without a Figma reference. ClickUp MCP unavailable during this run. Grounded in diff analysis + UX-expert evaluation.
> ℹ️ **Scope:** Governance section migration from Apollo/GraphQL to Kubb REST SDK. Most changes are hook-layer internals; findings below cover places where rendered output or UX behavior actually changes.

---

### 🔴 Must-fix

#### 1. `VotesTabContent.tsx` is only partially migrated — still imports from GraphQL client `[Code-only]`

**File:** `apps/dashboard/features/governance/components/proposal-overview/VotesTabContent.tsx`

While `useGetVotesQuery` and `useGetProposalNonVotersQuery` were replaced, this file still imports:
- `useTokenMetricsQuery` from `@anticapture/graphql-client/hooks`
- `OrderDirection` and `QueryInput_TokenMetrics_MetricType` from `@anticapture/graphql-client`

If the GraphQL client package is removed or deprecated (the direction this PR set out to achieve), this component will break silently at runtime. The `useTokenMetricsQuery` dependency should be migrated to the Kubb SDK equivalent before shipping.

---

### 🟡 Questions for author

#### 2. Search sends `" "` (single space) when field is cleared `[Code-only]`

**File:** `apps/dashboard/features/governance/components/governance-overview/GovernanceSection.tsx`

```diff
-variables: { query: trimmedSearch, limit: 50 },
-skip: !isSearchActive,
+{ query: trimmedSearch || " ", limit: 50 },
+{ query: { enabled: isSearchActive } },
```

When `trimmedSearch` is empty, the REST hook now receives `" "` (a space) rather than the query being blocked via `skip`. Does the REST `searchProposals` endpoint treat a single-space query identically to "no query"? If not, users who clear the search box may see a filtered or empty result set instead of the full proposals list.

#### 3. `calldatas` default changed from `null` to `[]` for offchain proposals `[Code-only]`

**File:** `apps/dashboard/features/governance/components/proposal-overview/ProposalSection.tsx`

```diff
-calldatas: null,
+calldatas: [],
```

Any downstream component that null-checks `calldatas` to show/hide a calldata-review UI section (e.g. the calldata viewer tab) will now always see a truthy array for offchain proposals. Please confirm no UI sections are unintentionally revealed for offchain proposals after this change.

---

### 🔵 Note for author

#### 4. `"Untitled Proposal"` fallback inconsistent between list and search paths `[Code-only]`

- `GovernanceSection.tsx` inline `toGovernanceProposal` ✓: falls back to `"Untitled Proposal"` when `proposal.title` is absent
- `useProposals.ts` list hook ✗: passes `proposal.title` directly with no fallback

If the API returns a null/undefined title (e.g. for a draft or partially indexed proposal), the list view renders a blank title while the search view renders `"Untitled Proposal"`. Fix: add the fallback in `useProposals.ts`:

```ts
title: proposal.title ?? "Untitled Proposal",
```

#### 5. `useAccountPower` now requires two sequential queries before showing user data `[UX-heuristic]`

**File:** `apps/dashboard/features/governance/hooks/useAccountPower.ts`

Previously `useGetAccountPowerQuery` fetched `votingPowerByAccountId` + `votesByProposalId` in a single round-trip. Now they are two independent REST queries, and `isLoading` is `votingPowerQuery.isLoading || votesQuery.isLoading`. The `ProposalHeader` / `ProposalItem` stay in loading state until both resolve. There's also an early-return guard `if (!votingPowerQuery.data) return null` which means users see nothing (rather than a skeleton) if the voting power query resolves faster than the votes query. Confirm the loading/skeleton state correctly covers both queries resolving independently.

#### 6. Voting power variation column may flash empty on new infinite-scroll pages `[UX-heuristic]`

**File:** `apps/dashboard/features/governance/hooks/useVotes.ts`

The `VotingPowerVariation` data is now fetched via a separate `queryClient.fetchQuery` call after each infinite-scroll page arrives. Until that secondary fetch resolves, the variation column renders empty for the newly loaded rows. The prior Apollo approach merged variations in the same request batch so the column appeared atomically with the vote rows. Consider rendering a skeleton or `—` in the variation cells while the secondary fetch is in-flight, to avoid the jarring empty→populated shift.

---

### ✅ Validated correct — no change needed

- `hasVoted` logic: `!!data.votesByProposalId` (truthy object check) → `(votesQuery.data?.items.length ?? 0) > 0` (explicit count) — the new logic is more correct and closes a potential "voted but shows unvoted" display bug. ✓
- `voterAddress.toLowerCase()` for power-map lookup — normalisation improvement, prevents case-mismatch misses. ✓
- Sort/filter state lifted to URL params (`sort`, `dir`, `support`, `voter`, `orderDirection`) via nuqs — UX improvement, state survives refresh and is shareable. ✓
- `OngoingProposalBanner` data path: `data?.proposals?.items` → `data?.items` — structurally equivalent, no visual change. ✓
- `transformToGovernanceProposal.ts` deleted, logic inlined — no visual change. ✓
- `proposals.graphql` removed — client-side cleanup only, no rendering impact. ✓
- No `globals.css` or token-level changes — no risk of global visual regression. ✓

---

_Review by @isadorable-png · automated UI review_

---

PR #1931, opened 2026-05-21
fix: issues on revenue and add token suggestions

brunod-e:
## Summary

Token suggestion chips on the Create Proposal transfer modal, plus a Revenue
dashboard overhaul: window-based KPIs, weighted Renewal Rate, YTD comparison,
and chart refinements.

## What changed

**Create Proposal**
- Suggested-token chips on `AddTransferModal` — click to populate address + decimals

**Revenue dashboard**
- **KpiRow** — `3M / 6M / 1Y / 3Y / MAX` presets with previous-period delta on
  each card; adaptive precision (e.g. `+0.4pp`, `+9pp`); MAX renders a
  `since {firstDate}` snapshot without delta
- **Renewal Rate** is now weighted by `termsExpiring` (was unweighted avg of
  monthly rates) — same fix applied to the "Renewal Rate by Expiry Year" cohort
- **Hero** gains a YTD comparison vs the same period last year, intersecting
  completed months in both years so it's robust to indexer lag and the
  in-progress current month
- **Name Growth & Churn** — single sign-colored bar per month (gain green /
  loss red, no more half-width split); Active Names snapshot moved here as a
  blue subtitle
- **Upcoming Expirations** — 7-year horizon; `X never-renewed names expiring
  in next 12 months` subtitle with calendar-aware boundaries (no off-by-one
  when current month rolls over); no `Total` row in tooltip
- **Monthly Revenue** — dual y-axis; tooltip shows per-stream values + `Total`
- **New Wallets** — info tooltip clarifying what counts as "interacting"

## Plumbing
- `transform.ts` split into `transform/charts.ts` + `transform/kpis.ts` +
  `utils/window.ts` (window math, delta presentation helpers)
- `StackedBarChart` tooltip is HTML-escaped; `Total` row is opt-in via the new
  `tooltipTotalLabel` prop
- All chart transforms wrapped in `useMemo`
- `formatMillions` fixed for values under $50K (was rendering literal `"$M"`)

## Test plan
- [ ] Token chips on Transfer modal populate address + decimals correctly
- [ ] KpiRow time presets actually change displayed values and deltas
- [ ] Spot-check the Renewal Rate KPI against a manual `sum(renewed)/sum(expiring)`
- [ ] All chart tooltips render and the Total row appears only on Monthly Revenue


pikonha (approved):

isadorable-png (approved):
Looks really good.
Small improvement that could be done: add a "X" at the end of the input so the user can "clear" the token that was populated in cased they want to paste something new.

PR #1936, opened 2026-05-26
Footer dashboard version

pikonha:
> [!NOTE]
> **Low Risk**
> Display and client-side query gating only; no auth, data writes, or critical path changes.
> 
> **Overview**
> **Footer version** now comes from the latest GitHub release whose tag starts with `dashboard@`, not the monorepo’s single “latest” release (which could be another package). The hook lists releases (up to 100), picks the first matching tag, and exposes a a normalized `version` string; the footer reads `release?.version` instead of `tag_name`.
> 
> **Proposal details** only run the SDK query when `proposalId` is truthy via React Query’s `enabled: !!proposalId`, avoiding fetches with an empty id.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bc7fc0aea4aca97869a62d08423c41e61790a6ab. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changeset/fix-footer-dashboard-version.md
+++ b/.changeset/fix-footer-dashboard-version.md
@@ -1,0 +1,5 @@
+---
+"@anticapture/dashboard": patch
+---
+
+Footer now shows the latest dashboard release version instead of the most recent repo release (which could belong to another package like gateful).
--- a/apps/dashboard/features/governance/hooks/useProposal.ts
+++ b/apps/dashboard/features/governance/hooks/useProposal.ts
@@ -21,7 +21,12 @@ export const useProposal = ({
   daoId,
 }: UseProposalParams): UseProposalResult => {
   const daoKey = daoId.toLowerCase() as ProposalPathParamsDaoEnumKey;
-  const { data, isLoading, error } = useProposalSDK(daoKey, proposalId, {});
+  const { data, isLoading, error } = useProposalSDK(
+    daoKey,
+    proposalId,
+    {},
+    { query: { enabled: !!proposalId } },
+  );
 
   return {
     data: data
--- a/apps/dashboard/shared/components/design-system/footer/Footer.tsx
+++ b/apps/dashboard/shared/components/design-system/footer/Footer.tsx
@@ -29,7 +29,7 @@ export type FooterProps = VariantProps<typeof footerVariant> & {
 
 export const Footer = ({ variant, className }: FooterProps) => {
   const { data: release } = useGitHubRelease();
-  const version = release?.tag_name || "v1.1.0";
+  const version = release?.version || "v1.1.0";
   const releaseUrl =
     release?.html_url || "https://github.com/blockful/anticapture/releases";
 
--- a/apps/dashboard/shared/hooks/useGitHubRelease.ts
+++ b/apps/dashboard/shared/hooks/useGitHubRelease.ts
@@ -8,18 +8,33 @@ interface GitHubRelease {
   published_at: string;
 }
 
-const GITHUB_API_URL =
-  "https://api.github.com/repos/blockful/anticapture/releases/latest";
+interface DashboardRelease extends GitHubRelease {
+  version: string;
+}
 
-export const fetchLatestRelease = async (): Promise<GitHubRelease> => {
-  const response = await axios.get(GITHUB_API_URL);
-  return response.data;
-};
+const GITHUB_API_URL =
+  "https://api.github.com/repos/blockful/anticapture/releases";
+const DASHBOARD_TAG_PREFIX = "dashboard@";
+
+export const fetchLatestDashboardRelease =
+  async (): Promise<DashboardRelease | null> => {
+    const response = await axios.get<GitHubRelease[]>(GITHUB_API_URL, {
+      params: { per_page: 100 },
+    });
+    const dashboardRelease = response.data.find((release) =>
+      release.tag_name.startsWith(DASHBOARD_TAG_PREFIX),
+    );
+    if (!dashboardRelease) return null;
+    return {
+      ...dashboardRelease,
+      version: `v${dashboardRelease.tag_name.slice(DASHBOARD_TAG_PREFIX.length)}`,
+    };
+  };
 
 export const useGitHubRelease = () => {
-  return useQuery<GitHubRelease, Error>({
-    queryKey: ["github-latest-release"],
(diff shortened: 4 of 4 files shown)


pikonha (commented):

Two issues worth fixing before merge — one is a latent bug, one is a structural clarity problem. The `useProposal` change is fine.

> **Verdict: blocking** — items 1 and 2 should be addressed before merging.

---

### 1. Fragile version normalization — silent `vv1.x.x` bug

**`useGitHubRelease.ts`**

```ts
version: `v${dashboardRelease.tag_name.slice(DASHBOARD_TAG_PREFIX.length)}`,
```

This unconditionally prepends `v`. If the tag is `dashboard@1.2.3` → `v1.2.3` ✓. If the tag is ever `dashboard@v1.2.3` → `vv1.2.3` ✗ — silently wrong in the footer with no error.

The fix is trivial:

```ts
const raw = dashboardRelease.tag_name.slice(DASHBOARD_TAG_PREFIX.length);
version: raw.startsWith("v") ? raw : `v${raw}`,
```

---

### 2. `DashboardRelease` leaks `tag_name` when `version` is the canonical field — missed code-judo simplification

`DashboardRelease` extends `GitHubRelease` via spread, so the returned object carries both `tag_name` and `version`. `Footer` only consumes `version` and `html_url`. Having both on the same object is confusing: which is the display version? `tag_name` is now an internal implementation detail but it's silently exposed to consumers.

The cleaner move: return only what consumers actually need from the hook, and keep the GitHub API shape entirely internal. This also removes the exported `fetchLatestDashboardRelease` (it's only a `queryFn` — nothing outside the hook needs it):

```ts
// Remove DashboardRelease interface entirely.
// Inline the fetch logic and expose a minimal typed result.

export const useGitHubRelease = () => {
  return useQuery<{ version: string; html_url: string } | null, Error>({
    queryKey: ["github-latest-dashboard-release"],
    queryFn: async () => {
      const response = await axios.get<GitHubRelease[]>(GITHUB_API_URL, {
        params: { per_page: 100 },
      });
      const match = response.data.find((r) =>
        r.tag_name.startsWith(DASHBOARD_TAG_PREFIX),
      );
      if (!match) return null;
      const raw = match.tag_name.slice(DASHBOARD_TAG_PREFIX.length);
      return {
        version: raw.startsWith("v") ? raw : `v${raw}`,
        html_url: match.html_url,
      };
    },
    staleTime: 3600000,
    gcTime: 3600000,
    refetchOnWindowFocus: false,
    refetchOnReconnect: false,
  });
};
```

This deletes `DashboardRelease`, removes the now-unnecessary `fetchLatestDashboardRelease` export, and makes the return type an explicit minimal contract. The GitHub API shape stays fully internal.

---

### 3. Minor: `proposalId: string` type lies about what it guards

**`useProposal.ts`**

```ts
export interface UseProposalParams {
  proposalId: string;  // typed as always-valid string
}

// but then:
{ query: { enabled: !!proposalId } }  // guards against empty string
```

The type says `proposalId` is always a non-null string, but the guard is handling the empty-string case (the component renders before the ID resolves). The type should reflect reality:

```ts
proposalId: string | undefined;
```

This makes the guard meaningful at the type level and signals to callers that passing `undefined` is an expected state. Non-blocking on its own, but it obscures the actual invariant and the fix is one word.

---

### What's good

- The `enabled: !!proposalId` fix itself is correct — prevents unnecessary fetches with an empty ID.
- Scoping to `dashboard@` tags is the right approach; using `.find()` on `per_page: 100` is fine given GitHub API constraints.
- Changeset included. ✓

---

pikonha (commented):

The `enabled: !!proposalId` guard and the changeset are clean and correct. The footer fix solves the real problem. One structural issue before merge, plus two non-blocking observations.

> **Note:** This review would normally be submitted as REQUEST_CHANGES — the blocking item below should be addressed before merging.

---

### [BLOCKING] `DashboardRelease extends GitHubRelease` leaks an implementation-specific `tag_name` format through the hook's public API

```ts
// useGitHubRelease.ts
interface DashboardRelease extends GitHubRelease {
  version: string;
}
```

Because `DashboardRelease` extends `GitHubRelease`, the hook returns an object where `tag_name` is `"dashboard@1.2.3"` and `version` is `"v1.2.3"` — side by side on the same object. The raw `tag_name` in this context has a monorepo-specific, implementation-internal format that no consumer should ever use for display or logic.

`Footer.tsx` only reads `release?.version` and `release?.html_url`. `tag_name`, `name`, and `published_at` are never consumed. But any future caller that reaches for `release?.tag_name` will silently get `"dashboard@1.2.3"` — the wrong format — with no type error to stop them.

The fix is direct: replace the `extends` approach with a clean, consumer-oriented interface that exposes only what callers actually need:

```ts
interface DashboardRelease {
  version: string;
  html_url: string;
}

export const fetchLatestDashboardRelease =
  async (): Promise<DashboardRelease | null> => {
    const response = await axios.get<GitHubRelease[]>(GITHUB_API_URL, {
      params: { per_page: 100 },
    });
    const release = response.data.find((r) =>
      r.tag_name.startsWith(DASHBOARD_TAG_PREFIX),
    );
    if (!release) return null;
    return {
      version: `v${release.tag_name.slice(DASHBOARD_TAG_PREFIX.length)}`,
      html_url: release.html_url,
    };
  };
```

This makes the contract explicit, removes the ambiguous `tag_name` from the public API, and prevents the silent misuse path. The transformation stays in the same place — it's a three-line change.

---

### [Non-blocking] Silent failure at >100 releases

`per_page: 100` is a hard cap. If the repo accumulates 100+ releases before a `dashboard@` tag appears in the list, `fetchLatestDashboardRelease` returns `null` and the footer silently falls back to the hardcoded `"v1.1.0"` string. Acceptable at current repo scale, but worth a comment noting this assumption so future debuggers aren't confused.

---

### [Non-blocking] Hidden version format invariant in the normalization

```ts
version: `v${dashboardRelease.tag_name.slice(DASHBOARD_TAG_PREFIX.length)}`
```

This assumes the tag suffix is bare semver (`1.2.3`), not `v`-prefixed (`v1.2.3`). A tag like `dashboard@v2.0.0` would produce `"vv2.0.0"`. If the tagging convention is enforced externally (e.g., by the release workflow), this is fine — but it's a hidden invariant that a comment or a defensive normalize step would make explicit.

---

PR #1941, opened 2026-05-26
feat: migrate attack-profitability hooks to kubb-generated REST SDK

brunod-e:


pikonha (changes requested):
## Code Quality Review — PR #1941

Overall direction is solid: the migration removes a lot of boilerplate and the resulting code is easier to follow. Two things need to be confirmed/fixed before merging.

---

### 🔴 Blocking

**1. `DaysWindow` mapping removed without verification — potential silent regression**

The old `useTreasury` did:
```ts
const daysKey = days as keyof typeof DaysWindow;
variables: { days: DaysWindow[daysKey], orderDirection: order === "asc" ? OrderDirection.Asc : OrderDirection.Desc }
```

The new code passes `days` (a `TimeInterval` string like `"365d"`) directly to the REST endpoint:
```ts
const params = { days, orderDirection: order };
```

If the REST endpoint expects a numeric value (`365`) or a different format than raw `TimeInterval` strings, this will silently return wrong/empty data — no TypeScript error, no runtime throw, just bad responses.

**Action required:** Confirm that the kubb-generated `useGetLiquidTreasury` / `useGetDaoTokenTreasury` / `useGetTotalTreasury` accept `"365d"`-style strings for `days` as defined by the REST API spec. If not, the mapping needs to be preserved.

---

**2. Unsafe `toLowerCase()` cast in both hooks**

```ts
// useDaoTokenHistoricalData.ts
const dao = daoId.toLowerCase() as HistoricalTokenDataPathParamsDaoEnumKey;

// useTreasury.ts
const dao = daoId.toLowerCase() as GetLiquidTreasuryPathParamsDaoEnumKey;
```

These are unchecked casts at the TypeScript level. If a `DaoIdEnum` value is ever added that doesn't exist in the REST API enum (or is spelled differently after lowercasing), the call fails at runtime with zero compile-time signal.

Also note: `useTreasury` casts to `GetLiquidTreasuryPathParamsDaoEnumKey` but then passes `dao` to all three hooks (`useGetDaoTokenTreasury`, `useGetTotalTreasury` too) — if those hooks have their own distinct path-param types, the cast is structurally wrong.

**Action required:** Either:
- Add a runtime guard / assertion, or
- Confirm all three hooks share the same path-param type and add a comment explaining why `GetLiquidTreasuryPathParamsDaoEnumKey` is used as the common cast.

---

### 🟡 Non-blocking

**3. `NO_CACHE_QUERY_OPTIONS` duplicated across both files**

Defined identically in `useDaoTokenHistoricalData.ts` and `useTreasury.ts`. Extract to a shared location (e.g., `apps/dashboard/features/attack-profitability/hooks/queryOptions.ts` or the feature's `utils/`) to prevent future drift between the two.

**4. `refetchOnMount: "always"` is redundant with `staleTime: 0`**

With `staleTime: 0`, React Query already treats all cached data as immediately stale and refetches on mount. `refetchOnMount: "always"` does the same thing. Not wrong — just adds noise to the "mirrors no-cache" comment.

---

### ✅ What's good

- Net -41 lines is a real win; the new code is much easier to reason about.
- Correct React hooks usage: all three hooks are always called, gated by `enabled` rather than conditionally invoked.
- The `active` ternary cleanly replaces the fragile `extractTreasuryData` type-narrowing chain.
- Changeset correctly scoped to `@anticapture/dashboard` patch.
- `null` filter removal from `useDaoTokenHistoricalData` is fine for REST (arrays don't carry nullable items the same way GraphQL does).

---

apps/dashboard/features/attack-profitability/hooks/useDaoTokenHistoricalData.ts
```diff
@@ -1,17 +1,14 @@
...
 
-type HistoricalTokenDataItem = NonNullable<
-  NonNullable<HistoricalTokenDataQuery["historicalTokenData"]>[number]
->;
-
-const toPriceEntry = (item: HistoricalTokenDataItem): PriceEntry => ({
+const toPriceEntry = (item: TokenHistoricalPriceItem): PriceEntry => ({
   price: item.price,
   timestamp: item.timestamp,
 });
```
pikonha: is this even needed?

apps/dashboard/features/attack-profitability/hooks/useTreasury.ts
```diff
@@ -19,81 +18,32 @@ export interface TreasuryDataPoint {
...
-  const treasury = extractTreasuryData(type, data);
+  const active =
+    type === "liquid" ? liquid : type === "dao-token" ? daoToken : total;
 
   return {
-    data: treasury?.items ?? [],
-    totalCount: treasury?.totalCount ?? 0,
-    loading,
-    error,
+    data: (active.data?.items ?? []) as TreasuryItem[],
```
pikonha: is this cast needed? couldn't this be type-narrowed?

apps/dashboard/features/attack-profitability/hooks/useDaoTokenHistoricalData.ts
pikonha: fixed

PR #1942, opened 2026-05-27
fix(dashboard): support Snapshot copeland votes and readable offchain choice labels

alextnetto:
## Summary

Fixes DEV-734 for ENS proposal **[6.44] Path forward on Working Groups for Term 7**.

## Root cause

1. **Empty vote modal** — The proposal uses Snapshot voting type **`copeland`** (Condorcet ranked ballot). `OffchainVotingModal` only handled standard Snapshot types, so `copeland` hit the `default` branch and rendered **no vote options**.
2. **Truncated labels** — Current Results used a fixed `w-[100px]` column with `truncate`, cutting ~32-character Snapshot labels.

Verified on Snapshot Hub: proposal `0xe4e1c052b2ea4f640cab27ddec326df6290d8996a9219b60cda4c4d4509f5f9a` has `type: "copeland"` and five choices. Votes use ranked arrays (e.g. `[5,4,3,2,1]`).

## Fix

- Map `copeland` to ranked-choice UI and sign votes as `ranked-choice` (required by snapshot.js for array payloads).
- Widen offchain result labels with tooltip for full text.
- Snapshot fallback link for unsupported types.

## Testing

- `offchainVotingType.test.ts`
- Dashboard typecheck (pre-push)

[Slack Thread](https://blockful.slack.com/archives/C08SL8B3GUF/p1779848782788729?thread_ts=1779848782.788729&cid=C08SL8B3GUF)

<div><a href="https://cursor.com/agents/bc-5a385556-b0c3-5d37-8986-c9028ab4f3ae"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-5a385556-b0c3-5d37-8986-c9028ab4f3ae"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/.changeset/copeland-offchain-voting.md
+++ b/.changeset/copeland-offchain-voting.md
@@ -1,0 +1,5 @@
+---
+"@anticapture/dashboard": patch
+---
+
+Support Snapshot copeland offchain votes and show full choice labels in proposal results.
--- a/apps/dashboard/features/governance/components/modals/OffchainVotingModal.tsx
+++ b/apps/dashboard/features/governance/components/modals/OffchainVotingModal.tsx
@@ -14,6 +14,7 @@ import { WeightedVoteOptions } from "@/features/governance/components/modals/vot
 import { useOffchainVotingPower } from "@/features/governance/hooks/useOffchainVotingPower";
 import { useVoteOnOffchainProposal } from "@/features/governance/hooks/useVoteOnOffchainProposal";
 import { normalizeChoices } from "@/features/governance/utils/offchainProposal";
+import { getOffchainVoteUiType } from "@/features/governance/utils/offchainVotingType";
 import { showCustomToast } from "@/features/governance/utils/showCustomToast";
 import { Button } from "@/shared/components";
 import { formatNumberUserReadable } from "@/shared/utils";
@@ -71,6 +72,7 @@ export const OffchainVotingModal = ({
   });
 
   const choices = normalizeChoices(proposal.choices);
+  const voteUiType = getOffchainVoteUiType(proposal.type);
 
   // Reset state when modal opens
   useEffect(() => {
@@ -112,15 +114,15 @@ export const OffchainVotingModal = ({
 
   const isVoteEnabled = (() => {
     if (!value || !address || isVoting) return false;
-    if (proposal.type === "approval") return (value as number[]).length > 0;
-    if (proposal.type === "weighted") {
+    if (voteUiType === "approval") return (value as number[]).length > 0;
+    if (voteUiType === "weighted") {
       const total = Object.values(value as Record<string, number>).reduce(
         (a, b) => a + b,
         0,
       );
       return total === 100;
     }
-    if (proposal.type === "quadratic") {
+    if (voteUiType === "quadratic") {
       const total = Object.values(value as Record<string, number>).reduce(
         (a, b) => a + b,
         0,
@@ -153,13 +155,7 @@ export const OffchainVotingModal = ({
       await vote({
         spaceId: proposal.spaceId,
         proposalId: proposal.id,
-        type: proposal.type as
-          | "basic"
-          | "single-choice"
-          | "approval"
-          | "ranked-choice"
-          | "weighted"
-          | "quadratic",
+        proposalType: proposal.type,
         choice: value,
         reason: comment,
       });
@@ -174,7 +170,7 @@ export const OffchainVotingModal = ({
   };
 
   const renderVoteOptions = () => {
-    switch (proposal.type) {
+    switch (voteUiType) {
       case "basic":
         return (
           <BasicVoteOptions
@@ -227,7 +223,21 @@ export const OffchainVotingModal = ({
           />
         );
       default:
-        return null;
+        return (
+          <p className="text-secondary text-[14px]">
+            Voting for &quot;{proposal.type}&quot; proposals is not supported in
+            this app yet. Vote on{" "}
+            <a
+              href={proposal.link}
+              target="_blank"
+              rel="noopener noreferrer"
+              className="text-link underline"
+            >
+              Snapshot
+            </a>
+            .
+          </p>
+        );
     }
   };
 
--- a/apps/dashboard/features/governance/components/modals/OffchainVotingModal.tsx
+++ b/apps/dashboard/features/governance/components/modals/OffchainVotingModal.tsx
@@ -301,11 +301,11 @@ export const OffchainVotingModal = ({
         </div>
 
         {/* Vote options */}
-        <div className="flex flex-col gap-[6px] p-4">
+        <div className="flex flex-col items-start gap-[6px] p-4 text-left">
           <p className="font-inter text-primary text-[12px] font-medium">
             Your vote
           </p>
-          {renderVoteOptions()}
+          <div className="w-full">{renderVoteOptions()}</div>
         </div>
 
         {/* Comment */}
--- a/apps/dashboard/features/governance/components/modals/vote-options/RankedChoiceOptions.tsx
+++ b/apps/dashboard/features/governance/components/modals/vote-options/RankedChoiceOptions.tsx
@@ -47,13 +47,13 @@ export const RankedChoiceOptions = ({
         <div
           key={choiceIndex}
           className={cn(
-            "border-border-default flex items-center gap-2 border px-[10px] py-2",
+            "border-border-default flex items-start gap-2 border px-[10px] py-2 text-left",
           )}
         >
-          <span className="text-secondary font-inter w-5 shrink-0 text-[14px] font-normal not-italic leading-[20px]">
+          <span className="text-secondary font-inter w-5 shrink-0 text-left text-[14px] font-normal not-italic leading-[20px]">
             {position + 1}
           </span>
-          <span className="font-inter text-primary flex-1 text-[14px] font-normal not-italic leading-[20px]">
+          <span className="font-inter text-primary min-w-0 flex-1 text-left text-[14px] font-normal not-italic leading-[20px]">
             {choices[choiceIndex]}
           </span>
           <div className="flex flex-col">
--- a/apps/dashboard/features/governance/components/proposal-overview/OffchainVoteLabelChip.tsx
+++ b/apps/dashboard/features/governance/components/proposal-overview/OffchainVoteLabelChip.tsx
@@ -1,0 +1,53 @@
+"use client";
+
+import { presentOffchainVoteLabel } from "@/features/governance/utils/offchainVoteLabel";
+import { Tooltip } from "@/shared/components/design-system/tooltips/Tooltip";
+import { cn } from "@/shared/utils";
+
+interface OffchainVoteLabelChipProps {
+  label: string;
+  proposalType?: string | null;
+  className?: string;
+  variant?: "badge" | "inline";
+}
+
+export const OffchainVoteLabelChip = ({
+  label,
+  proposalType,
+  className,
+  variant = "badge",
+}: OffchainVoteLabelChipProps) => {
+  const { display, full, showTooltip } = presentOffchainVoteLabel(
+    label,
+    proposalType,
+  );
+
+  const chip = (
+    <span
+      className={cn(
+        "text-primary font-inter truncate text-left font-medium not-italic",
+        variant === "badge"
+          ? "bg-surface-default max-w-[200px] rounded-full px-[6px] py-[2px] text-[12px] leading-[16px]"
+          : "max-w-[140px] text-sm leading-[20px]",
+        className,
+      )}
+    >
+      {display}
+    </span>
+  );
+
+  if (!showTooltip) {
+    return chip;
+  }
+
+  return (
+    <Tooltip
+      asChild
+      triggerClassName="max-w-[200px]"
+      tooltipContent={full}
+      className="text-left"
+    >
+      {chip}
+    </Tooltip>
+  );
+};
--- a/apps/dashboard/features/governance/components/proposal-overview/OffchainVotesContent.tsx
+++ b/apps/dashboard/features/governance/components/proposal-overview/OffchainVotesContent.tsx
@@ -11,7 +11,9 @@ import { CheckCircle2, CircleMinus, Inbox, XCircle } from "lucide-react";
 import { useCallback, useEffect, useMemo, useRef, useState } from "react";
 import type { Address } from "viem";
 
+import { OffchainVoteLabelChip } from "@/features/governance/components/proposal-overview/OffchainVoteLabelChip";
 import { VotesTable } from "@/features/governance/components/proposal-overview/VotesTable";
+import { getOffchainVoteFullLabel } from "@/features/governance/utils/offchainVoteLabel";
 import { BlankSlate } from "@/shared/components/design-system/blank-slate/BlankSlate";
 import { Button } from "@/shared/components/design-system/buttons/button/Button";
 import { SkeletonRow } from "@/shared/components/skeletons/SkeletonRow";
@@ -38,20 +40,14 @@ const getLabelDisplay = (label: string) => {
 const getChoiceInfo = (choice: string[], choices: string[]) => {
   if (choice.length === 0) return { label: "—", icon: null as React.ReactNode };
 
+  const fullLabel = getOffchainVoteFullLabel(choice, choices);
+  if (!fullLabel) return { label: "—", icon: null as React.ReactNode };
+
   if (choice.length === 1 && choice[0] != null) {
-    const idx = Number(choice[0]);
-    const label = choices[idx - 1] ?? `Choice ${choice[0]}`;
-    return getLabelDisplay(label);
+    return getLabelDisplay(fullLabel);
   }
 
-  const label = choice
-    .filter((c): c is string => c != null)
-    .map((c) => {
-      const idx = Number(c);
-      return choices[idx - 1] ?? `Choice ${c}`;
-    })
-    .join(", ");
-  return { label, icon: null as React.ReactNode };
+  return { label: fullLabel, icon: null as React.ReactNode };
 };
 
 interface OffchainVotesContentProps {
@@ -59,6 +55,7 @@ interface OffchainVotesContentProps {
   daoId: DaoIdEnum;
   totalVotingPower: number;
   choices: string[];
+  proposalType?: string | null;
 }
 
 export const OffchainVotesContent = ({
@@ -66,6 +63,7 @@ export const OffchainVotesContent = ({
   daoId,
   totalVotingPower,
   choices,
+  proposalType,
 }: OffchainVotesContentProps) => {
   const loadingRowRef = useRef<HTMLTableRowElement>(null);
 
@@ -270,11 +268,21 @@ export const OffchainVotesContent = ({
             );
           }
           const choice = row.getValue("choice") as string[];
-          const { icon, label } = getChoiceInfo(choice, choices);
+          const choiceInfo = getChoiceInfo(choice, choices);
+          const { icon, label } = choiceInfo;
+          const isMultiChoice = choice.length > 1;
           return (
-            <div className="flex items-center gap-2 p-2">
+            <div className="flex min-w-0 items-center gap-2 p-2">
               {icon}
-              <span className="text-sm font-medium">{label}</span>
+              {isMultiChoice ? (
+                <OffchainVoteLabelChip
+                  label={label}
+                  proposalType={proposalType}
+                  variant="inline"
+                />
+              ) : (
+                <span className="text-sm font-medium">{label}</span>
+              )}
             </div>
           );
         },
@@ -401,7 +409,14 @@ export const OffchainVotesContent = ({
         },
       },
     ],
-    [totalVotingPower, choices, orderBy, orderDirection, handleSort],
+    [
+      totalVotingPower,
+      choices,
+      proposalType,
+      orderBy,
+      orderDirection,
+      handleSort,
+    ],
   );
 
   if (error) return <div>Error: {error.message}</div>;
--- a/apps/dashboard/features/governance/components/proposal-overview/ProposalHeader.tsx
+++ b/apps/dashboard/features/governance/components/proposal-overview/ProposalHeader.tsx
@@ -6,6 +6,7 @@ import Link from "next/link";
 import { usePathname } from "next/navigation";
 import type { Address } from "viem";
 
+import { OffchainVoteLabelChip } from "@/features/governance/components/proposal-overview/OffchainVoteLabelChip";
 import { BadgeStatus, Button } from "@/shared/components";
 import { ConnectWalletCustom } from "@/shared/components/wallet/ConnectWalletCustom";
 import { WhitelabelConnectWallet } from "@/shared/components/wallet/WhitelabelConnectWallet";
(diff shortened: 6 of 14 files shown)


pikonha:
@cursoragent make the options alligned left
image

pikonha:
@cursoragent align the square with the option text

image

pikonha:
@cursoragent remove the gap-1.5 between the option and the progress bar

pikonha (changes requested):

The fix is conceptually right — mapping `copeland` to the ranked-choice UI/signing path is clean, the fallback message is an improvement over silent `null`, and the layout work is fine. Two structural issues in `offchainVotingType.ts` need addressing before merge.

---

### 1. `RANKED_BALLOT_TYPES` is a redundant sync point — delete it

`RANKED_BALLOT_TYPES = new Set(["ranked-choice", "copeland"])` encodes the exact same invariant that `UI_TYPE_ALIASES = { copeland: "ranked-choice" }` already encodes. "This type uses ranked ballots" is equivalent to "this type's UI type is ranked-choice", and `getOffchainVoteUiType` is already the canonical resolver for that.

Both consumers of `RANKED_BALLOT_TYPES` can be derived from `getOffchainVoteUiType` instead:

```ts
// current — has to be kept in sync with UI_TYPE_ALIASES manually
export const getSnapshotVoteSignType = (proposalType: string): OffchainVoteUiType => {
  if (RANKED_BALLOT_TYPES.has(proposalType)) return "ranked-choice";   // redundant
  return getOffchainVoteUiType(proposalType) ?? "single-choice";
};

export const usesRankedBallot = (proposalType: string): boolean =>
  RANKED_BALLOT_TYPES.has(proposalType);                               // redundant
```

```ts
// proposed — single source of truth
export const getSnapshotVoteSignType = (proposalType: string): OffchainVoteUiType =>
  getOffchainVoteUiType(proposalType) ?? "single-choice";

export const usesRankedBallot = (proposalType: string): boolean =>
  getOffchainVoteUiType(proposalType) === "ranked-choice";
```

When the next Condorcet method (e.g. `schulze`) comes in, the current code requires adding it to **both** `UI_TYPE_ALIASES` and `RANKED_BALLOT_TYPES`. With the proposed form, adding it to `UI_TYPE_ALIASES` alone is sufficient and correct — no second place to forget.

---

### 2. `getSnapshotVoteSignType` return type should use `OffchainVoteUiType`, not a re-spelled union

The return type is:

```ts
): "basic" | "single-choice" | "approval" | "ranked-choice" | "weighted" | "quadratic"
```

This is the exact definition of `OffchainVoteUiType`, which is already exported from the same file. Use the type:

```ts
): OffchainVoteUiType
```

Same fix applies if `usesRankedBallot`'s signature becomes the target of a future refactor. Keep one authoritative definition.

---

### 3. `usesRankedBallot` is exported but appears unused in component code

It's tested but not referenced in `OffchainVotingModal` or `useVoteOnOffchainProposal`. If it's pre-shipped for a caller that doesn't exist yet, remove it or keep it internal. Exporting dead surface area locks in the API before the usage is clear.

---

pikonha (changes requested):
## Code Quality Review

The fix direction is right — extracting vote-type logic into its own module, removing the cast, adding tests, showing a Snapshot fallback for unknown types. That part is good. But `offchainVotingType.ts` ships with a structural redundancy that should be cut before this merges.

---

### [BLOCKING] `RANKED_BALLOT_TYPES` and `UI_TYPE_ALIASES` are two sources of truth for the same relationship

`offchainVotingType.ts` currently has two parallel structures that both encode "copeland is a ranked-choice ballot":

```ts
const RANKED_BALLOT_TYPES = new Set(["ranked-choice", "copeland"]);  // encodes it
const UI_TYPE_ALIASES: Record<string, OffchainVoteUiType> = {
  copeland: "ranked-choice",  // also encodes it
};
```

When the next ranked-ballot alias comes along, someone has to remember to update both. That's the definition of a maintainability trap.

`getSnapshotVoteSignType` compounds this: it first does a `RANKED_BALLOT_TYPES.has(proposalType)` early-exit, then falls back to `getOffchainVoteUiType`. But `getOffchainVoteUiType` already handles `copeland → ranked-choice` via `UI_TYPE_ALIASES`, so the early-exit is entirely redundant — removing it doesn't change any output.

The code-judo move here: **delete `RANKED_BALLOT_TYPES` entirely** and collapse `getSnapshotVoteSignType` to a one-liner that reuses the function you already have:

```ts
export const getSnapshotVoteSignType = (proposalType: string): OffchainVoteUiType =>
  getOffchainVoteUiType(proposalType) ?? "single-choice";
```

Behavior is identical for all inputs. ~14 lines gone. One concept instead of two.

---

### [BLOCKING] `usesRankedBallot` is dead production code

`usesRankedBallot` is exported and has a test, but it has no callers in any production file touched by this PR. Exporting untested surface is noise at best, confusion at worst (readers will wonder where it's called from). 

If it genuinely has no callers, delete it. If it's intended for a future caller, add it when that caller lands.

---

### [MINOR] `getSnapshotVoteSignType` inlines `OffchainVoteUiType` instead of reusing it

```ts
export const getSnapshotVoteSignType = (
  proposalType: string,
):
  | "basic"
  | "single-choice"
  | "approval"
  | "ranked-choice"
  | "weighted"
  | "quadratic" => {
```

`OffchainVoteUiType` is already exported from the same file. The return type should use it directly. This is inconsistency, and if `OffchainVoteUiType` ever gains a new member the return type here silently diverges.

---

### [MINOR] `isVoteEnabled` relies on implicit null guard

In `OffchainVotingModal`, when `voteUiType` is `null` (unsupported proposal type), `isVoteEnabled` falls through all the if-checks and returns `!!value`. The button ends up disabled only because `renderVoteOptions` returns the Snapshot link instead of a vote component, meaning `value` is never set.

This is a fragile implicit contract. Adding one explicit line makes the invariant obvious and prevents a future vote-options change from accidentally enabling the button for unsupported types:

```ts
const isVoteEnabled = (() => {
  if (!value || !address || isVoting) return false;
  if (voteUiType === null) return false;  // unsupported type — no vote options rendered
  if (voteUiType === "approval") return (value as number[]).length > 0;
  // ...
```

---

### What's good

- Pulling vote-type logic into a dedicated utility is the right call. The `OffchainVotingModal` is meaningfully cleaner for it.
- Removing the `proposal.type as SnapshotProposalType` cast and passing `proposalType: string` through to `getSnapshotVoteSignType` is strictly better.
- The Snapshot fallback link for unsupported types is the right UX.
- The label layout change in `ProposalInfoSection` is a reasonable structural trade-off for long labels.
- Tests are present and cover the key alias behavior.

The two blocking items are both in `offchainVotingType.ts` and resolve to the same root cause: `RANKED_BALLOT_TYPES` should not exist. Deleting it and its consumer (`usesRankedBallot`) reduces the file from ~60 lines to ~45 and removes a whole class of "update two things" bugs.


---

pikonha (changes requested):
## Code Quality Review

The copeland fix itself is clean and well-structured (`offchainVotingType.ts` + tests are solid). However the PR also bundles a full GraphQL → REST migration across ~10 hooks, a new API feature, and multiple unrelated bug fixes. I found two blocking bugs and one performance regression introduced in the migration portion.

---

### 🔴 Blocking — `useAccountInteractionsData`: totalCount uses last page instead of first

**File:** `apps/dashboard/features/holders-and-delegates/token-holder/drawer/top-interactions/hooks/useAccountInteractionsData.ts`

```ts
const totalCount = data?.pages[data.pages.length - 1]?.totalCount ?? 0;
```

The PR description explicitly documents this exact bug for `useProposalsActivity`:

> _"The summary fields (`totalProposals`, etc.) are only returned by the API on the first page, so the next-page check now anchors on the **first page's total** instead of the **last page's** (which was always 0 after the initial fetch)."_

`useProposalsActivity` was correctly fixed to use `allPages[0]?.totalProposals`. But `useAccountInteractionsData` still reads from the **last** page. When a second page loads, the last page's `totalCount` is 0, which makes this guard immediately trigger:

```ts
if (!topFive.length || totalCount === 0) {
  return defaultData; // ← empty state shown when paginating
}
```

The top-interactions chart/table would wipe itself clean as soon as the user scrolls to load more. Fix:

```ts
// use first page — totalCount is only present on page 1
const totalCount = data?.pages[0]?.totalCount ?? 0;
```

---

### 🔴 Blocking — `useDelegators`: dedup set not memoized

**File:** `apps/dashboard/shared/hooks/useDelegators.ts`

```ts
const seen = new Set<string>();
const delegators =
  data?.pages
    .flatMap((p) => p.items)
    .filter((item) => { ... }) ?? [];
```

This `flatMap` + `filter` runs on **every render**, not only when `data` changes. Every parent re-render (e.g. a scroll event, a hover, a sibling state update) re-executes the full traversal of all loaded pages.

All other refactored hooks in this same PR wrap the identical pattern in `useMemo` — `useDelegationHistory`, `useAccountInteractionsData`, `useProposalsActivity`, `useDelegateDelegationHistory`. This one is the odd one out:

```ts
// consistent with the rest of the PR
const delegators = useMemo(() => {
  const seen = new Set<string>();
  return (
    data?.pages
      .flatMap((p) => p.items)
      .filter((item) => {
        if (seen.has(item.delegatorAddress)) return false;
        seen.add(item.delegatorAddress);
        return true;
      }) ?? []
  );
}, [data]);
```

---

### 🟡 Non-blocking — `fetchNextPageStable` wrapper is unnecessary

**File:** `apps/dashboard/features/holders-and-delegates/hooks/useDelegates.ts`

```ts
const fetchNextPageStable = useCallback(async () => {
  await fetchNextPage();
}, [fetchNextPage]);
```

`fetchNextPage` from TanStack Query's `useInfiniteQuery` is already a stable reference (guaranteed by the library). This wrapper adds indirection with no benefit and can be removed.

---

### 🟡 Non-blocking — PR scope vs. title

The title is `fix(dashboard): support Snapshot copeland votes and readable offchain choice labels` but the diff includes:

- Full GraphQL → REST migration for 10+ hooks (`useBalanceHistory`, `useDelegates`, `useDelegationHistory`, `useProposalsActivity`, `useDelegateDelegationHistory`, `useDelegators`, `useAccountInteractionsData`, …)
- New API feature (`from`/`to` filter params on `/balances/historical`)
- Infinite-scroll fix for proposals activity
- Activity-feed E2E test fix
- `BLOCKFUL_API_TOKEN` env var addition

Each of these could break independently and is hard to bisect. Ideally the migration lives in a separate PR. At minimum the PR description and changeset name should reflect the true surface area.

---

### Summary

| # | Severity | File | Issue |
|---|----------|------|-------|
| 1 | 🔴 Bug | `useAccountInteractionsData.ts` | `totalCount` from last page zeroes out on scroll |
| 2 | 🔴 Perf regression | `useDelegators.ts` | Dedup flatMap outside `useMemo` |
| 3 | 🟡 Nit | `useDelegates.ts` | Unnecessary `fetchNextPageStable` wrapper |
| 4 | 🟡 Process | multiple | PR mixes unrelated concerns |


---

pikonha (changes requested):
## Code Review

Read through all 4 032 lines of diff. Existing comments (options alignment, square alignment) were already addressed — not repeated here.

---

### Blocking

**1. `useDelegators.ts` — `seen` Set is not memoized (React anti-pattern)**

```ts
// apps/dashboard/shared/hooks/useDelegators.ts
const seen = new Set<string>();          // recreated every render
const delegators =
  data?.pages
    .flatMap((p) => p.items)
    .filter((item) => {
      if (seen.has(item.delegatorAddress)) return false;
      seen.add(item.delegatorAddress);   // mutates during render
      return true;
    }) ?? [];
```

This is the only hook in the PR that does deduplication outside of `useMemo`. Mutating state during render is a React rule violation and will double-execute in Strict Mode (dev), producing unexpected results. Every other migrated hook (`useDelegationHistory`, `useProposalsActivity`, `useAccountInteractionsData`) wraps the same pattern in `useMemo`. Fix:

```ts
const delegators = useMemo(() => {
  const seen = new Set<string>();
  return (
    data?.pages
      .flatMap((p) => p.items)
      .filter((item) => {
        if (seen.has(item.delegatorAddress)) return false;
        seen.add(item.delegatorAddress);
        return true;
      }) ?? []
  );
}, [data]);
```

---

**2. Graph hooks hardcode `limit: 1000` — silent data truncation**

`useBalanceHistoryGraph.ts` and `useDelegateDelegationHistoryGraph.ts` both added:

```ts
fromValue: "1",
limit: 1000,
```

For a delegate or holder with more than 1 000 historical events the graph will silently show an incomplete picture with no error or warning. A 1 000-item cap may be fine as a pragmatic choice, but it should either:
- Paginate in the background (fetch all pages) and merge, or
- Log a warning / display a notice when `totalCount > 1000`, so the user knows data is clipped.

The `fromValue: "1"` filter (excluding zero-delta rows) is also an undocumented behaviour change — worth a comment explaining why.

---

### Non-blocking observations

**`fetchNextPageStable` wrapper adds no value**

```ts
// useDelegates.ts
const fetchNextPageStable = useCallback(async () => {
  await fetchNextPage();
}, [fetchNextPage]);
```

React Query's `fetchNextPage` is already stable across renders. This wrapper is dead code. Remove it and return `fetchNextPage` directly.

---

**`getSnapshotVoteSignType` silently falls back to `"single-choice"` for unknown types**

```ts
// offchainVotingType.ts
return "single-choice"; // reached for any type the UI doesn't know about
```

The modal's `renderVoteOptions` returns a Snapshot link and blocks voting for unknown types, so this path shouldn't be reachable in practice. Still, adding a `console.warn` or a `never`-style assertion makes the invariant explicit and easier to debug if a new type slips through.

---

**`userVoteFilter` in `DelegateProposalsActivity` changed from `useState` to `useQueryState`**

This persists the vote filter in the URL, which is a visible behaviour change: sharing/bookmarking a URL will pre-select the filter. If that's intentional it should be noted; if not, revert to `useState`.

---

**`useBalanceHistoryGraph` renamed its public return shape**

`{ balanceHistory, loading }` → `{ data, isLoading }`. The consuming component was updated correctly, but if any other caller (outside this PR) destructures the old names it will silently get `undefined`. Worth a search or a quick `grep balanceHistory` across the codebase to confirm there are no other consumers.

---

**Scope / changeset**

The PR bundles five distinct changesets:
1. Copeland/offchain voting fix
2. Balance-history `from`/`to` server-side filters
3. Proposals-activity infinite-scroll fix
4. Holders-and-delegates REST migration
5. Vote-composition delegators REST migration

The copeland fix is small and self-contained; the REST migration touches ~35 files. Bundling makes bisecting regressions harder. If these can't be split now, that's fine, but worth noting for future PRs.

---

### What's well done

- `offchainVotingType.ts` is cleanly abstracted, tested, and correctly separates UI rendering type from EIP-712 signing type.
- The `copeland → ranked-choice` mapping and fallback Snapshot link for unsupported types are the right approach.
- The `getProposalsNextPage` anchor on `allPages[0]?.totalProposals` correctly fixes the infinite scroll bug described in the changeset.
- Integration tests for the new `from`/`to` filter parameters are thorough.
- The GraphQL `fetchMore` + manual page-tracking pattern has been consistently replaced with React Query's `useInfiniteQuery`, which is a meaningful cleanup.


---

pikonha (changes requested):

The copeland fix itself is clean and well-structured. The REST migration is the substantial bulk of this PR — most of it is a genuine improvement (deleting ~600 lines of manual pagination boilerplate, removing Apollo dependency from these hooks). A few things need addressing before merge.

---

### 🔴 BLOCKING

#### 1. `getHistoricalBalanceCount` has two structurally diverging query paths

`apps/api/src/repositories/account-balance/historical.ts`

The count method now branches:
- `!fromAddress && !toAddress` → `this.db.$count(balanceHistory, ...)` — **no join to `transfer`**
- address filter present → inner join with `transfer` → `count(*)::int`

The items query always joins `transfer` (it has to — it returns transfer metadata). This means if any `balanceHistory` row exists without a matching `transfer` row, the count path will over-count: `totalCount > actual items returned`. The UI would show a "load more" button that returns empty results.

The code-judo move here: **remove the `if (!fromAddress && !toAddress)` branch entirely and always use the joined count query.** The join is on indexed columns (`transactionHash`, `logIndex`), not meaningfully more expensive than `$count()`. The result: one code path, correct by construction, no divergence risk.

```ts
// BEFORE: two diverging paths
async getHistoricalBalanceCount(...): Promise<number> {
  if (!fromAddress && !toAddress) {
    return await this.db.$count(balanceHistory, and(...)); // no join
  }
  const [row] = await this.db
    .select({ count: sql<number>`count(*)::int` })
    .from(balanceHistory)
    .innerJoin(transfer, ...)
    .where(and(...));
  return row?.count ?? 0;
}

// AFTER: single unified path
async getHistoricalBalanceCount(...): Promise<number> {
  const [row] = await this.db
    .select({ count: sql<number>`count(*)::int` })
    .from(balanceHistory)
    .innerJoin(transfer, ...)
    .where(and(...));
  return row?.count ?? 0;
}
```

---

### 🟡 STRUCTURAL (should fix before merge)

#### 2. `useDelegators.ts` — dedup runs on every render

`apps/dashboard/shared/hooks/useDelegators.ts`

```ts
const seen = new Set<string>();
const delegators =
  data?.pages
    .flatMap((p) => p.items)
    .filter((item) => {
      if (seen.has(item.delegatorAddress)) return false;
      seen.add(item.delegatorAddress);
      return true;
    }) ?? [];
```

`seen` is recreated and the full array is recomputed on **every render**. Every other hook in this PR memoizes this pattern correctly — `useDelegationHistory`, `useProposalsActivity`, `useAccountInteractionsData` all use `useMemo`. This one doesn't. Wrap in `useMemo([data])`.

---

#### 3. `DelegationHistoryTable.tsx` — silent feature regression: sort-by-amount removed

`apps/dashboard/features/holders-and-delegates/token-holder/drawer/delegation-history/DelegationHistoryTable.tsx`

The `orderBy` param was silently dropped from `useDelegationHistory` and the `AmountFilter`'s sort options ("Largest first / Smallest first") were removed. Users can no longer sort delegation history by amount.

If the REST endpoint for historical delegations doesn't support `orderBy=amount`, that's a valid constraint — but the PR description and changeset don't mention this regression. Is this intentional? If so, it should be explicitly noted. If the API supports it, the param should be wired through.

---

### 🟠 DESIGN (clean up in this PR or follow-up)

#### 4. `offchainVotingType.ts` — `RANKED_BALLOT_TYPES` duplicates `UI_TYPE_ALIASES`

`apps/dashboard/features/governance/utils/offchainVotingType.ts`

`RANKED_BALLOT_TYPES = new Set(["ranked-choice", "copeland"])` and `UI_TYPE_ALIASES = { copeland: "ranked-choice" }` are encoding the same fact from two directions. `getSnapshotVoteSignType` checks the Set first, then falls through to `getOffchainVoteUiType`. This duplication will silently drift if a new alias is added to one but not the other.

The Set can be deleted entirely:

```ts
// No RANKED_BALLOT_TYPES needed

export const getSnapshotVoteSignType = (proposalType: string) =>
  getOffchainVoteUiType(proposalType) ?? "single-choice";

export const usesRankedBallot = (proposalType: string): boolean =>
  getOffchainVoteUiType(proposalType) === "ranked-choice";
```

`getOffchainVoteUiType` already normalizes aliases via `UI_TYPE_ALIASES`, so copeland → "ranked-choice" is already handled. One source of truth.

---

#### 5. `useAccountInteractionsData.ts` — `totalCount` anchored to last page, not first

```ts
const totalCount = data?.pages[data.pages.length - 1]?.totalCount ?? 0;
```

Every other hook in this PR anchors summary fields to `pages[0]` — `useProposalsActivity` uses `allPages[0]?.totalProposals` explicitly because summary fields come from the first page. Using the *last* page here is inconsistent and semantically odd: total count is a property of the full result set, not of whichever page was fetched most recently. Use `data.pages[0]?.totalCount`.

---

#### 6. Graph hooks — magic `limit: 1000` with no explanation

`useBalanceHistoryGraph.ts` and `useDelegateDelegationHistoryGraph.ts` both hardcode `limit: 1000`. If a delegate has >1000 history items the graph silently truncates. This is a behavioral regression from the old hooks (which didn't have this cap). At minimum, add a comment explaining the bound and acknowledging the truncation risk. Ideally wire through a configurable limit or use the infinite variant with `getNextPageParam`.

---

#### 7. `fetchNextPageStable` in `useDelegates.ts` — thin wrapper with no structural value

```ts
const fetchNextPageStable = useCallback(async () => {
  await fetchNextPage();
}, [fetchNextPage]);
```

`fetchNextPage` from TanStack Query's `useInfiniteQuery` is already referentially stable across renders. This wrapper exists only to coerce the return type from `Promise<InfiniteData<...>>` to `Promise<void>`. Either update the `UseDelegatesResult` interface to accept the TanStack return type, or cast: `fetchNextPage as () => Promise<void>`. The wrapper adds an unnecessary async frame.

---

### ✅ What's good

- The copeland alias approach in `offchainVotingType.ts` is the right model — alias map + switch is cleaner than scattering `|| proposal.type === "copeland"` checks across the modal.
- Deleting `PaginationInfo` / `SimplePaginationInfo` / `pagination` objects and replacing with a flat `hasNextPage: boolean` is a real simplification — the old pagination objects were exposing `currentPage`, `totalPages`, `currentItemsCount` that callers never used.
- Deleting `graphql-client/useDelegators.ts` (181 lines) and replacing with 65 lines in `useDelegators.ts` is a genuine code-judo win.
- The `getProposalsNextPage` anchoring to `allPages[0]?.totalProposals` is correct — that field is only present on the first page.
- Removing the manual `useEffect` reset-to-page-1 + `useState(currentPage)` + `updateQuery` accumulator patterns from every hook is the right call.


---

pikonha:
@cursoragent fix this

```
 🟠 DESIGN (clean up in this PR or follow-up)
4. offchainVotingType.ts — RANKED_BALLOT_TYPES duplicates UI_TYPE_ALIASES
apps/dashboard/features/governance/utils/offchainVotingType.ts

RANKED_BALLOT_TYPES = new Set(["ranked-choice", "copeland"]) and UI_TYPE_ALIASES = { copeland: "ranked-choice" } are encoding the same fact from two directions. getSnapshotVoteSignType checks the Set first, then falls through to getOffchainVoteUiType. This duplication will silently drift if a new alias is added to one but not the other.

The Set can be deleted entirely:

// No RANKED_BALLOT_TYPES needed

export const getSnapshotVoteSignType = (proposalType: string) =>
  getOffchainVoteUiType(proposalType) ?? "single-choice";

export const usesRankedBallot = (proposalType: string): boolean =>
  getOffchainVoteUiType(proposalType) === "ranked-choice";
getOffchainVoteUiType already normalizes aliases via UI_TYPE_ALIASES, so copeland → "ranked-choice" is already handled. One source of truth.
```

brunod-e (approved):

pikonha:
@cursoragent voting on these multiple options leads to this huge label. make it be shown on hover for these multiple options scnearios

image

PR #1944, opened 2026-05-27
fix: token hodlers pagination

pikonha:


pikonha (commented):

One blocking type regression before merge. The rest of the cleanup is solid.

> ⚠️ Note: GitHub prevents requesting changes on your own PR — would normally be REQUEST_CHANGES.

---

### 🔴 [Blocking] `Promise<unknown>` type regression in `useDelegates.ts`

Changing the interface from `Promise<void>` to `Promise<unknown>` is a step backward. `unknown` communicates nothing useful to callers and actively obscures the contract — it's worse than `void` for a value nobody consumes.

The root cause is correct: removing `fetchNextPageStable` (the identity `useCallback` wrapper) forces a type adjustment because React Query's `fetchNextPage` doesn't return `Promise<void>`. But `Promise<unknown>` is the wrong fix.

**Preferred remedy** — since no caller uses the return value, the honest contract is:

```ts
fetchNextPage: () => void;
```

If TypeScript complains at the return site, a narrow cast is the right answer:

```ts
fetchNextPage: fetchNextPage as () => void,
```

This makes the boundary explicit ("callers don't need this") instead of deferring with `unknown`.

---

### ✅ Good moves

**Removing `fetchNextPageStable`** — this was a textbook identity wrapper. `useCallback(async () => { await fn() }, [fn])` rememoizes only when `fn` changes, which is identical to just passing `fn`. Good deletion.

**`useMemo` in `useDelegators`** — the dedup `Set` was being reconstructed on every render. Dependency array `[data]` is correct. Clean fix.

**`totalCount` from `pages[0]`** — reading from the last loaded page was fragile; the total is stable and lives on every page response. Reading from page 0 is the right anchor.

**Passing `getNextPageParam` to `useAccountBalancesInfinite`** — the actual bug. Consistent with `useDelegators`. Correct.

---

### 🟡 [Non-blocking] API count query: confirm the general path handles the no-address case

The removed branch in `historical.ts` used `this.db.$count()` — Drizzle's native count with no join. The surviving general query almost certainly JOINs with the accounts table. Verify that:

1. The WHERE clause correctly omits address filtering when both `fromAddress`/`toAddress` are absent.
2. The JOIN itself isn't applied unconditionally — otherwise the common case (no address filter) now pays for a join it didn't before.

The simplification is defensible — one code path is better than two — but worth a quick confirmation that the general query is truly a superset, not a coincidentally working approximation.

---

### 📝 Minor: title typo

`fix: token hodlers pagination` → `fix: token holders pagination`

---

pikonha (commented):
## Code Review — `fix/token-holders-pagination`

Overall: solid, focused fix. All five changes are correct and hang together well. No blocking issues.

---

### `apps/api/src/repositories/account-balance/historical.ts`

The removed early-return branch used `this.db.$count(balanceHistory, ...)` — **no JOIN with the `transfer` table**. Meanwhile, `getHistoricalBalances` always does `INNER JOIN transfer`. This meant the count and the list were querying different datasets when no `fromAddress`/`toAddress` was given, so the reported total could be higher than the actual number of records returned (overcounting `balanceHistory` rows that have no matching `transfer` entry).

Removing the branch forces `getHistoricalBalanceCount` through the same `INNER JOIN transfer` path, making count and list results consistent. This is the kind of subtle inconsistency that shows up as off-by-N pagination bugs — good catch.

---

### `apps/dashboard/features/holders-and-delegates/hooks/useTokenHolders.ts`

Root cause of the broken pagination. `useAccountBalancesInfinite` was called without `getNextPageParam`, so React Query never knew what cursor to use for the next page and would always refetch page 1. Passing `{ query: { getNextPageParam } }` is the correct fix, consistent with how `useVotingPowersInfinite` and `useDelegatorsInfinite` are already called.

---

### `apps/dashboard/features/holders-and-delegates/hooks/useDelegates.ts`

`fetchNextPageStable` was a `useCallback(() => fetchNextPage())` with no transformation — React Query's `fetchNextPage` is already referentially stable, so the wrapper added zero value.

The type change from `Promise<void>` to `Promise<unknown>` is correct; `Promise<void>` was a lie in the interface. Worth a quick scan of callers to confirm none rely on the narrower type in a way TypeScript would now flag.

---

### `apps/dashboard/features/holders-and-delegates/token-holder/drawer/top-interactions/hooks/useAccountInteractionsData.ts`

`totalCount` represents the API's total item count, not a per-page value — it should be the same on every page. Using `pages[0]` is semantically cleaner (the total is known from the first response) and safer: `pages[0]` is always defined when `data.pages` has any entries, whereas `pages[pages.length - 1]` would return `undefined` if `pages` were somehow empty after the nullish check.

---

### `apps/dashboard/shared/hooks/useDelegators.ts`

The deduplication logic was running on every render. Wrapping it in `useMemo([data])` is the right fix — React Query keeps the `data` reference stable when nothing changes, so the memo will hit reliably.

---

### Minor nit

PR title has a typo: **"hodlers"** → "holders". Worth fixing for git history searchability.

---

pikonha (commented):
## Code Review — fix: token holders pagination

### What this PR does (and why the fixes are correct)

**Root cause (`useTokenHolders.ts`)** — Core fix is correct. `useAccountBalancesInfinite` was called without a `getNextPageParam`, so react-query had no way to derive the next-page cursor and infinite scroll silently stopped working. ✅

**`historical.ts`** — Fix is correct but needs documentation (see below). The removed `$count()` branch queried `balanceHistory` alone. The data query (`getHistoricalBalances`) does an `INNER JOIN` on `transfer`. Any `balanceHistory` row without a matching `transfer` row is included in `$count` but excluded from data — so the count was systematically higher than the actual row count, producing phantom extra pages. Making `getHistoricalBalanceCount` use the same `INNER JOIN` makes count and data consistent. ✅

**`useDelegators.ts` `useMemo`** — Correct. The deduplication block was recreating a `Set` and re-filtering on every render. `useMemo([data])` gates it on actual data changes. ✅

**`useAccountInteractionsData.ts` `pages[0]`** — Correct. `totalCount` is a property of the full dataset, not a per-page value. Reading it from `pages[0]` is more stable than `pages[pages.length - 1]` since the first page is always present. ✅

---

### 🔴 Blocking

**1. Type regression in `UseDelegatesResult`**

```ts
// Before
fetchNextPage: () => Promise<void>;

// After
fetchNextPage: () => Promise<unknown>;
```

`Promise<void>` is the correct contract here — it communicates "call this and discard the return." Widening to `Promise<unknown>` is semantically wrong: it implies the caller *should* inspect the result, and is not assignable to `Promise<void>` in TypeScript, which will break any caller typed against this interface:

```ts
// Breaks after this change:
const handler: () => Promise<void> = delegates.fetchNextPage;
```

If you want the type to reflect react-query's actual return, use the proper generic. Otherwise, keep `Promise<void>`. The removal of `fetchNextPageStable` is fine (react-query v5 already provides a stable reference) — just fix the return type in the interface.

**2. No explanation for the `historical.ts` behavioral change**

The `$count` path was over-counting because it lacked the `INNER JOIN transfer`. This is a non-obvious invariant: **the count query must mirror the data query's joins and filters exactly, or pagination breaks.** Without a comment or PR description explaining this, a future developer will see two queries doing seemingly the same thing and "optimize" by removing the join — reintroducing the bug.

Please add a short inline comment on the join, or describe the root cause in the PR description.

---

### 🟡 Non-blocking suggestions

**Missing test coverage** — No test verifies that `getHistoricalBalanceCount` and `getHistoricalBalances` return consistent row counts (the exact class of bug this PR fixes). Not blocking for a hotfix, but worth a follow-up.

**PR title typo** — `"fix: token hodlers pagination"` → `"fix: token holders pagination"`

---

### Summary

The pagination logic is correctly fixed across all five files. Two things should be addressed before merging: the `Promise<void>` → `Promise<unknown>` type regression in `UseDelegatesResult`, and a note (inline or in the PR description) explaining why `getHistoricalBalanceCount` must JOIN on `transfer`.

---

brunod-e (approved):

PR #1946, opened 2026-05-27
Dev

pikonha:


LeonardoVieira1630 (approved):

pikonha (commented):

The GraphQL → REST SDK migration is well-scoped and the deletions are genuinely good (killing all the hand-rolled `fetchMore` + `updateQuery` + page-state patterns). Most of the new hooks are cleaner. There are two blocking issues and four things that should be fixed before merge.

---

### 🔴 BLOCKING — logic bug: `presentOffchainVoteLabel` detects multi-choice by scanning the pre-joined string

**File:** `apps/dashboard/features/governance/utils/offchainVoteLabel.ts`

```ts
if (uiType === "ranked-choice") {
  const rankedChoices = fullLabel.split(", ");   // ← splits the joined string
  if (rankedChoices.length > 1) { ... }
}

if (uiType === "weighted" || uiType === "quadratic") {
  if (fullLabel.length > 40 || fullLabel.includes(", ")) { ... }
}
```

`fullLabel` is already the output of `getOffchainVoteFullLabel`, which joins choice names with `", "`. A single choice named `"3 WGs + secretary (As-is), option 2"` would split into 2 parts and falsely display as `"Ranked (2 choices)"`. A choice named `"50/50 split, rebalance"` would incorrectly render as `"Split vote"`.

Real governance proposals (ENS working-group votes, treasury allocation votes) routinely have commas in choice names.

**The root cause is `presentOffchainVoteLabel` takes only the pre-joined string, discarding the structure it actually needs.** The function should receive either:
- the raw `choice: string[]` array and `choices: string[]`, or
- at minimum the `choiceCount: number` alongside the joined label.

Then the multi-choice check is `choice.length > 1`, not `fullLabel.split(", ").length > 1`. The tooltip display still uses the full joined string — nothing else changes.

---

### 🔴 BLOCKING — three hooks re-implement the canonical `getNextPageParam` already exported from `@anticapture/client`

**Files:**
- `apps/dashboard/features/holders-and-delegates/hooks/useBalanceHistory.ts`
- `apps/dashboard/features/holders-and-delegates/hooks/useDelegateDelegationHistory.ts`
- `apps/dashboard/features/holders-and-delegates/hooks/useDelegationHistory.ts`

Each defines:

```ts
const getNextPageParam = (lastPage, allPages) => {
  const loaded = allPages.reduce((s, p) => s + p.items.length, 0);
  return loaded >= lastPage.totalCount ? undefined : loaded;
};
```

This is identical logic to the `getNextPageParam` already exported from `@anticapture/client`, which the other three hooks in this same PR (`useTokenHolders`, `useDelegators`, `useAccountInteractionsData`) already import and use correctly. The three local copies will drift if the canonical implementation is ever changed. This is the exact scenario the canonical utility exists to prevent.

If the TypeScript generic on the exported version doesn't accept the specific response types, the fix is either to generalize the exported signature or to cast — not to silently fork the implementation.

---

### 🟡 SHOULD FIX — `useBalanceHistoryGraph` renames its public interface inconsistently

The hook changed its return shape from `{ balanceHistory, loading, error }` to `{ data, isLoading, error }`. Every other hook in this PR kept `loading` (not `isLoading`). The consumer now has to destructure-rename:

```ts
const { data: balanceHistory, isLoading: loading, error } = useBalanceHistoryGraph(...)
```

Either keep the hook's original external names (`balanceHistory`, `loading`) or normalize *all* hooks in this file set to `isLoading`. Right now it's uniquely inconsistent, which will confuse the next person who adds a consumer.

---

### 🟡 SHOULD FIX — `StringifyBigInt<T>` in `useDelegateDelegationHistory` makes a promise the code doesn't keep

```ts
type StringifyBigInt<T> = {
  [K in keyof T]: T[K] extends RestBigInt ? string : T[K];
};

delegation: StringifyBigInt<NonNullable<HistoricalVotingPowerDelegation>> | null;
transfer:   StringifyBigInt<NonNullable<HistoricalVotingPowerTransfer>> | null;
```

The actual mapping only converts specific known fields (`.value`, `.votingPower`, `.timestamp`). If the REST type gains a new `BigInt` field, this type annotation says it's a `string` — but the runtime code won't have converted it. The type makes the wrong promise. Use explicit inline types for the two mapped shapes instead of a generic transformer that doesn't reflect the real conversion. That also makes the interface immediately readable without mentally applying the type transform.

---

### 🟡 SHOULD FIX — `hasPreviousPage` in `useDelegateDelegationHistory` is semantically meaningless

```ts
hasPreviousPage: (data?.pages?.length ?? 0) > 1,
```

This hook feeds an infinite-scroll table. There is no "go back" action in the UI. Using loaded page count as a proxy produces a value that becomes permanently `true` once page 2 is loaded, and is consumed nowhere downstream. It's part of the exported `UseDelegateDelegationHistoryResult` interface — remove it before someone writes code against it expecting it to mean something.

---

### 🟡 SHOULD FIX — dedup in four hooks raises an unanswered question about backend cursor correctness

`useDelegators`, `useDelegationHistory`, `useAccountInteractionsData`, and `useProposalsActivity` all implement the same `seen = new Set()` dedup after flattening pages. If the pagination offset cursor is correct, each page returns unique items by definition.

The dedup means one of two things: (a) there's a known backend issue where items can span page boundaries, or (b) this is purely defensive. Either way it should be documented in a comment explaining *why* the dedup is necessary, or the underlying issue should be fixed and the dedup removed. Unexplained dedup teaches future contributors that "of course you need to dedup" and proliferates the pattern into every new hook. Four occurrences in a single PR is the signal.

---

### Minor

- `accountId as Address` on the fallback path in `VotingPowerHistoryTable.tsx` (~line 908, `delegateAddress = accountId as Address`) — the caller passes `accountId: string` but `Address` is `0x${string}`. This cast was present before; worth fixing since the surrounding code was touched.
- `error as Error | undefined` cast in `useAccountInteractionsData.ts` — TanStack Query's `error` is already `Error | null`. Either adjust the `InteractionResponse` interface to use `Error | null` or remove the cast.


---

pikonha (commented):
## Code Quality Review

This is a significant and largely well-executed migration — deleting hundreds of lines of manual Apollo `fetchMore` + page-state boilerplate is a real improvement. The offchain voting type abstraction is clean and well-tested. Two structural issues need to be fixed before merge, plus several non-blocking concerns worth addressing.

---

### 🚫 Blocking

#### 1. Three hooks define their own `getNextPageParam` instead of importing the canonical one

`useBalanceHistory.ts`, `useDelegateDelegationHistory.ts`, and `useDelegationHistory.ts` each define a local `getNextPageParam`:

```ts
const getNextPageParam = (lastPage, allPages) => {
  const loaded = allPages.reduce((s, p) => s + p.items.length, 0);
  return loaded >= lastPage.totalCount ? undefined : loaded;
};
```

This same PR imports `getNextPageParam` from `@anticapture/client` in **four other hooks** — `useDelegators`, `useTokenHolders`, `useAccountInteractionsData`, and `useDelegates`. The canonical utility is right there, already being used in this very PR. The three local copies are redundant and will drift.

`useProposalsActivity.ts` is the one legitimate exception — its `getProposalsNextPage` anchors to `allPages[0]?.totalProposals` and counts `.proposals`, which is genuinely different.

Fix: replace the three local definitions with `import { getNextPageParam } from "@anticapture/client"`.

---

#### 2. `presentOffchainVoteLabel` re-parses a pre-formatted string to extract structure

```ts
// offchainVoteLabel.ts
if (uiType === "ranked-choice") {
  const rankedChoices = fullLabel.split(", ");
  if (rankedChoices.length > 1) {
    return { display: `Ranked (${rankedChoices.length} choices)`, ... };
  }
}
```

`presentOffchainVoteLabel` receives the already-joined string from `getOffchainVoteFullLabel` and splits on `", "` to count choices. This couples the presentation function to the join separator used in the formatting function — a silent, untested dependency. If the separator or format ever changes in `getOffchainVoteFullLabel`, `presentOffchainVoteLabel` will silently count wrong without failing tests.

The same split-to-count pattern appears for `approval` and `weighted`/`quadratic` too.

The fix is to pass the count (or the raw choices array) alongside the label so `presentOffchainVoteLabel` doesn't need to recover information that was already available before formatting:

```ts
// option: pass count explicitly
export const presentOffchainVoteLabel = (
  fullLabel: string,
  choiceCount: number,
  proposalType: string | null | undefined,
): OffchainVoteLabelPresentation
```

---

### ⚠️ Non-blocking concerns

#### 3. `daoId.toLowerCase() as DaoEnumKey` repeated 10+ times with no shared helper

Every migrated hook has:
```ts
daoId.toLowerCase() as HistoricalBalancesPathParamsDaoEnumKey
daoId.toLowerCase() as VotingPowersPathParamsDaoEnumKey
// etc.
```

This cast appears in at least 10 call sites. It's a shared operation with a shared invariant — a single typed cast utility would remove all of them. Not a blocker, but this is exactly the kind of repeated boilerplate that should live in one place.

#### 4. Client-side deduplication in four hooks papers over a potential pagination bug

`useDelegationHistory`, `useDelegators`, `useAccountInteractionsData`, and `useProposalsActivity` all deduplicate with a local `seen` set after `flatMap`-ing across pages:

```ts
const seen = new Set<string>();
return data?.pages.flatMap((p) => p.items).filter((item) => {
  if (seen.has(key)) return false;
  seen.add(key);
  return true;
}) ?? [];
```

React Query's `useInfiniteQuery` with a correct `getNextPageParam` (offset-based: return `loaded` as the next skip) should never produce duplicates unless there's a race condition or a server-side pagination bug. This dedup is a symptom, not a solution. Worth investigating whether the server actually returns duplicates before accepting this as the right fix in the client.

#### 5. `limit: 1000` truncation is not documented in the code

The changeset says _"document the single-page `limit: 1000` truncation in the balance-history and delegate-delegation-history graph hooks"_ but neither `useBalanceHistoryGraph.ts` nor `useDelegateDelegationHistoryGraph.ts` has any inline comment explaining why this is 1000, what happens when the real dataset exceeds it, or that this is an acknowledged trade-off. The changeset is not the right place for this — put it in the code.

#### 6. `useMemo` removed from `useBalanceHistoryGraph` data transformation

Old:
```ts
const balanceHistory = useMemo(() => {
  return data.items.map(...).sort(...);
}, [data, accountId, decimals]);
```

New (inline, no memo):
```ts
data: data?.items ? data.items.map(...).sort(...) : [],
```

The map+sort runs on every render now. For graph data this is potentially noticeable. Low-priority but should be a conscious choice.

#### 7. `offchainProposalType` drills through 3 component levels

`ProposalSection` → `ProposalHeader` → `ProposalHeaderAction` → `OffchainVotedBadge`, and also `ProposalSection` → `TabsSection` → `OffchainVotesContent`. The `ProposalHeaderContext` is already in scope here. This would be a natural home for `proposalType` rather than threading it as a prop through every intermediate component. Not a blocker but worth addressing before this prop drill gets worse as more types are added.

#### 8. Feature regression: amount sort removed from delegation history table

`DelegationHistoryTable.tsx` previously supported "Largest first" / "Smallest first" sort via the `AmountFilter`. The PR removes `setSortBy("amount")` from the `onApply` callback and removes `sortOptions` from the filter entirely. If the REST API doesn't support `orderBy=amount` for this endpoint, that's understandable, but it should be documented somewhere (changeset or a comment in the table).

---

### Summary

The migration overall is a clear improvement. Items 1 and 2 should be fixed before merge — both are straightforward.

---

pikonha (commented):

Overall the migration from GraphQL to the REST client is a clear win — the hooks shrink dramatically, pagination logic stops being hand-rolled, and the `updateQuery` spaghetti disappears. The offchain voting type abstraction is well-isolated. These are solid structural improvements.

Three things need to be fixed before merge; the rest are advisory.

---

### 🔴 BLOCKING

---

#### 1. `presentOffchainVoteLabel` round-trips through a lossy string and breaks on choice labels that contain `", "`

`getOffchainVoteFullLabel` collapses the structured `choice: string[]` into a comma-joined string. `presentOffchainVoteLabel` then splits that string back on `", "` to count how many choices were cast:

```ts
// offchainVoteLabel.ts
if (uiType === "ranked-choice") {
  const rankedChoices = fullLabel.split(", ");
  if (rankedChoices.length > 1) {
    return { display: `Ranked (${rankedChoices.length} choices)`, ... };
  }
}
```

If any choice label contains `", "` (e.g. `"Option A, Special"`) the split produces phantom extra entries — a 3-choice ranked vote would show `"Ranked (4 choices)"`.

The test passes only because the fixture choices happen not to contain `", "`.

**Fix:** pass the original cardinality instead of re-parsing the string.

Option A — extend the signature of `presentOffchainVoteLabel`:
```ts
export const presentOffchainVoteLabel = (
  fullLabel: string,
  choiceCount: number,
  proposalType: string | null | undefined,
): OffchainVoteLabelPresentation
```

Option B — keep structured data all the way to the chip and only stringify at render time. The `OffchainVoteLabelChip` already has a `proposalType` prop; adding a `choiceCount` prop would be just as cheap.

Either way, `split(", ")` should not be the thing that tells us how many choices a user cast.

---

#### 2. `getNextPageParam` is duplicated across three hooks instead of using the canonical export from `@anticapture/client`

`useTokenHolders`, `useDelegates`, `useAccountInteractionsData`, and `useDelegators` all do:
```ts
import { getNextPageParam } from "@anticapture/client";
```

Meanwhile `useBalanceHistory`, `useDelegateDelegationHistory`, and `useDelegationHistory` each define an identical local copy:
```ts
const getNextPageParam = (lastPage, allPages) => {
  const loaded = allPages.reduce((s, p) => s + p.items.length, 0);
  return loaded >= lastPage.totalCount ? undefined : loaded;
};
```

The body is character-for-character the same. Import from `@anticapture/client` in those three hooks too. If the exported function is typed too narrowly to accept those response types, widen the export's generic constraint — don't maintain three clones.

---

#### 3. Unsafe `error as Error | undefined` cast in `useAccountInteractionsData`

```ts
error: error as Error | undefined,
```

React Query's `error` is `Error | null`. Casting with `as` instead of narrowing loses null-safety. Replace with:

```ts
error: error ?? undefined,
```

This appears twice in the file (`defaultData` and the final return).

---

### 🟡 ADVISORY (non-blocking)

---

#### 4. `getOffchainVoteFullLabel` has a redundant single-element branch

```ts
if (choice.length === 1 && choice[0] != null) {
  const idx = Number(choice[0]);
  return choices[idx - 1] ?? `Choice ${choice[0]}`;
}
```

A single-element array joined with `", "` is just the element, so this branch is dead weight. Drop it:

```ts
if (choice.length === 0) return "";
return choice
  .filter((c): c is string => c != null)
  .map((c) => choices[Number(c) - 1] ?? `Choice ${c}`)
  .join(", ");
```

---

#### 5. `useBalanceHistoryGraph` exposes `isLoading` while every sibling hook remaps to `loading`

All other migrated hooks do `return { loading: isLoading, ... }`. This hook skips the remap and exposes `isLoading` directly. The caller was updated, but the inconsistency makes the layer's public shape harder to scan. Consider aligning.

---

#### 6. `hasPreviousPage` is dead weight in `UseDelegateDelegationHistoryResult`

```ts
hasPreviousPage: (data?.pages?.length ?? 0) > 1,
```

No caller reads this field after the infinite-scroll migration. Remove it to make clear the hook is forward-only.

---

#### 7. The dedup `seen Set` pattern appears in four hooks — worth a sanity check

`useDelegators`, `useDelegationHistory`, `useAccountInteractionsData`, and `useProposalsActivity` all filter out duplicate items via a `seen` set after flattening pages. If `getNextPageParam` is returning the correct offset, the server shouldn't return duplicates. Either document why duplicates are expected (e.g. a race window), or confirm the guards can be removed once pagination is verified correct.

---

The blocking items are all localized and fixable without touching the migration strategy.

---

pikonha (commented):
## Code Quality Review

Scope: all 57 changed files across API, Gateful, Gateway, and Dashboard. No prior review comments exist; this review covers everything from scratch.

---

### 🔴 Blocking

#### 1. `offchainVoteLabel.ts` — comma-split logic is fragile for choice labels that contain `, `

`presentOffchainVoteLabel` detects ranked/approval/multi-choice votes by splitting the `fullLabel` string on `", "`:

```ts
// apps/dashboard/features/governance/utils/offchainVoteLabel.ts
const rankedChoices = fullLabel.split(", ");
if (rankedChoices.length > 1) { ... }
```

`fullLabel` is itself built by `getOffchainVoteFullLabel` joining choice text with `", "`. Any proposal whose choices contain a literal `, ` in their label (e.g. `"Option A, funding $100k"`) will be mis-classified as a multi-choice vote, showing "Ranked (2 choices)" with a tooltip even though the voter picked a single option.

**Fix:** track the number of choices separately rather than re-parsing the already-joined string. For example, return `{ display, full, showTooltip, choiceCount }` from `getOffchainVoteFullLabel`, or pass the raw `choice` array down to `presentOffchainVoteLabel`.

The unit tests only cover clean, comma-free labels — add a case like `"Option A, the standard"` as a single copeland choice to cover this.

---

#### 2. `historical.ts` repository — `innerJoin` in count vs join type in data query

The new `getHistoricalBalanceCount` uses an `INNER JOIN` with the `transfer` table:

```ts
// apps/api/src/repositories/account-balance/historical.ts
const [row] = await this.db
  .select({ count: sql<number>`count(*)::int` })
  .from(balanceHistory)
  .innerJoin(transfer, sql`...`)
  .where(...)
```

The original data query (`getHistoricalBalances`) already joined with `transfer` (evidenced by `DBHistoricalBalanceWithRelations` returning transfer fields). If that join is a `LEFT JOIN`, then `totalCount` and `items.length` will diverge whenever a `balanceHistory` row has no matching `transfer` row — the data query includes the orphan record, the count query excludes it. This manifests as pagination cutting off early and the last page appearing "empty."

Please confirm the join type in the data query and ensure both queries are consistent.

---

### 🟡 Non-blocking but worth addressing

#### 3. Duplicate local `getNextPageParam` implementations

`useTokenHolders`, `useDelegators`, `useDelegates`, and `useAccountInteractionsData` all import the shared `getNextPageParam` from `@anticapture/client`. But `useBalanceHistory`, `useDelegateDelegationHistory`, `useDelegationHistory`, and `useProposalsActivity` each define their own local version with subtly different logic (different field names, different anchor strategies). If the shared helper covers the same contract, replace the duplicates with the import. Duplicate logic drifts over time.

---

#### 4. `DelegationHistoryTable` — silent removal of amount-based sorting

The `sortOptions` prop and the amount sort wiring in `AmountFilter` were removed along with the `orderBy` parameter from `useDelegationHistory`. The changeset entry for this PR doesn't mention this behaviour change. If intentional (REST endpoint doesn't support `orderBy` for delegations), add a note to the changeset. If accidental, restore it.

---

#### 5. `useBalanceHistoryGraph` / `useDelegateDelegationHistoryGraph` — `limit: 1000` silent truncation

```ts
{ fromValue: "1", limit: 1000, ... }
```

Both graph hooks silently discard data beyond 1000 items with no indication to the consumer. For addresses with long history the charts will be silently incomplete. The changeset mentions this limitation but it remains invisible to callers. At minimum return an `isTruncated` flag, or emit a console warning when `data.items.length === 1000`.

---

#### 6. `DelegateProposalsActivity` — `userVoteFilter` promoted to URL param

```ts
-  const [userVoteFilter, setUserVoteFilter] = useState<string>("all");
+  const [userVoteFilter, setUserVoteFilter] = useQueryState("userVoteFilter", ...)
```

Promoting to a URL param is generally better (survives navigation, shareable), but verify the param name `userVoteFilter` doesn't collide with any other query param on the same page. Also the `void setUserVoteFilter(...)` suppresses the floating-promise correctly but a comment or `await` would make the intent clearer.

---

#### 7. `useDelegates.ts` — loading state suppressed after first fetch

```ts
loading: isLoading && !allDelegates.length,
```

After the first successful fetch, `isLoading` will never surface even on refetch or filter change — skeletons won't appear on subsequent fetches. Prefer `isLoading` directly, or `isFetching` if you want to distinguish initial vs background refresh.

---

#### 8. E2E test — assertion now requires events to exist (potential CI flakiness)

```ts
// apps/dashboard/e2e/activity-feed.spec.ts
- await expect(hasFeedItems.or(hasMediumItems).or(isEmpty).or(failedToLoad)).toBeVisible(...)
+ await expect(eventRelevanceBadge).toBeVisible({ timeout: 30_000 });
```

The old test accepted an empty feed; the new one requires at least one event with High or Medium relevance. If the test environment has no indexed data this will fail. The original fallbacks were deliberate. Restore them or add a data precondition.

---

#### 9. Minor — `|| 0` vs `?? 0`

Several places use `|| 0` where `?? 0` is more precise (avoids treating a genuine server-returned `0` as falsy), e.g. `useDelegators.ts` line `totalCount: data?.pages[0]?.totalCount || 0` and similar in `useAccountInteractionsData.ts`.

---

### ✅ What looks good

- The `from`/`to` filter addition is clean end-to-end (mapper → service → repository → controller → OpenAPI → client) with solid integration + unit tests.
- The `UI_TYPE_ALIASES` map pattern in `offchainVotingType.ts` is extensible and well-tested.
- Replacing Apollo `fetchMore` + `updateQuery` manual merge boilerplate with `useXxxInfinite` is a meaningful correctness improvement.
- Anchoring `totalCount` to the first page in `useProposalsActivity` correctly fixes the infinite-scroll bug.
- `isLoadingMore={fetchingMore}` fixes across all tables are correct — `loading` was the wrong signal there.
- `BLOCKFUL_API_TOKEN` wired consistently into CI and Turbo env passlists.

---

pikonha (commented):
# Thermo-Nuclear Code Quality Review

> ⚠️ **Would be REQUEST_CHANGES** — three blocking issues listed below. Posting as a comment since author = reviewer.

Overall the migration from Apollo/GraphQL to the REST SDK is a solid cleanup — removing ~1200 lines of hand-rolled `fetchMore` / pagination state is exactly the right move and most of it lands well. A few structural issues need to be addressed before merge.

---

## 🚫 Blocking

### 1. `getNextPageParam` is duplicated across 3 hooks when the canonical export already exists

`@anticapture/client` already exports `getNextPageParam`. Four hooks (`useDelegates`, `useDelegators`, `useTokenHolders`, `useAccountInteractionsData`) correctly import it from there. But three hooks define the **identical function locally**:

- `useBalanceHistory.ts`
- `useDelegateDelegationHistory.ts`
- `useDelegationHistory.ts`

All three implement:
```ts
const getNextPageParam = (lastPage, allPages) => {
  const loaded = allPages.reduce((s, p) => s + p.items.length, 0);
  return loaded >= lastPage.totalCount ? undefined : loaded;
};
```

This is a bespoke helper for something we already have a canonical utility for. Drop the local definitions and import from `@anticapture/client` like the other hooks do.

(`useProposalsActivity`'s `getProposalsNextPage` is the intentional exception — it pivots on `allPages[0]?.totalProposals` because summary fields only return on the first page. That one is correct as-is.)

---

### 2. Silent behavior regression: amount-based `orderBy` silently removed from `DelegationHistoryTable`

The old table let users sort by `amount` via the amount filter's sort option. The new version drops the `sortBy` query-state, removes `sortOptions` from `<AmountFilter>`, and the `useDelegationHistory` interface no longer accepts `orderBy` at all.

If this is intentional (the REST endpoint doesn't support `orderBy=amount` for delegations), it needs to be:
- noted explicitly in the changeset, and
- the removed UI explained in code (or a TODO if the endpoint will support it later).

Right now users silently lose a sort option with no explanation anywhere in the code or changeset.

---

### 3. `limit: 1000` hard-cap in graph hooks is undocumented in-code

Both `useBalanceHistoryGraph` and `useDelegateDelegationHistoryGraph` now request up to 1000 items in a single non-paginated call:

```ts
fromValue: "1",
limit: 1000,
```

The changeset mentions the truncation, but there's no comment in the source. Users with >1000 entries will get silently truncated graphs. Add a comment explaining the known truncation and ideally link to a follow-up ticket. "Document the truncation" means **in the code**, not just in the changeset.

---

## ⚠️ Major Quality Issues

### 4. Magic numbers `40`/`37` in `offchainVoteLabel.ts` + fragile comma-split detection

```ts
if (uiType === "weighted" || uiType === "quadratic") {
  if (fullLabel.length > 40 || fullLabel.includes(", ")) {
    return { display: "Split vote", ... };
  }
}
// ...
if (fullLabel.length > 40) {
  return { display: `${fullLabel.slice(0, 37)}…`, ... };
}
```

Two issues:

**a)** `40` and `37` are unexplained magic constants. Extract them:
```ts
const LABEL_TRUNCATE_THRESHOLD = 40;
const LABEL_TRUNCATE_LENGTH = 37; // threshold - "…".length
```

**b)** `fullLabel.includes(", ")` detects multi-choice by checking for the join separator used in `getOffchainVoteFullLabel`. This is a fragile implicit contract — a single-choice label like `"Option A, B or C"` would falsely trigger "Split vote". The multi-choice check should operate on `choice.length > 1` before joining, not on the joined string's contents.

---

### 5. `onUserVoteFilterChange` prop type is `string`, forcing an `as` cast at the call site

In `ProposalsTable.tsx`:
```ts
onUserVoteFilterChange?: (filter: string) => void;
```

In `DelegateProposalsActivity.tsx`:
```tsx
onUserVoteFilterChange={(filter) => {
  void setUserVoteFilter(
    filter as "all" | "yes" | "no" | "abstain" | "no_vote",
  );
}}
```

The `as` cast papers over a too-wide prop type. Tighten `onUserVoteFilterChange` to accept the actual enum type and drop the cast.

---

### 6. `useBalanceHistoryGraph` returns `isLoading` while every other hook returns `loading`

```ts
// hook return:
return { data, isLoading, error };

// every call site:
const { data: balanceHistory, isLoading: loading, error } = useBalanceHistoryGraph(...);
```

Every consumer has to alias `isLoading: loading`. Rename the return key to `loading` to match the rest of the codebase.

---

## 🔍 Minor / Nits

### 7. `computedHasNextPage` alias in `useAccountInteractionsData.ts` is pointless

```ts
const computedHasNextPage = hasNextPage ?? false;
// used once:
hasNextPage: computedHasNextPage,
```

Inline it: `hasNextPage: hasNextPage ?? false`.

---

### 8. `StringifyBigInt<T>` adds type indirection without real enforcement

```ts
type StringifyBigInt<T> = {
  [K in keyof T]: T[K] extends RestBigInt ? string : T[K];
};
```

Used on `DelegationHistoryItem.delegation` and `.transfer`. The actual conversion is done manually in the `map()` with `.toString()` calls — the type doesn't catch a forgotten `.toString()`. Consider just declaring the interface fields directly with `string` types instead of the mapped generic.

---

### 9. `OffchainVoteLabelChip.tsx` — tooltip `triggerClassName` duplicates the chip's `max-w`

```tsx
<Tooltip triggerClassName="max-w-[200px]">
  <span className="... max-w-[200px] ...">
```

If the chip's `max-w` changes, the trigger won't follow automatically. Derive one from the other or use a shared constant.

---

## Summary

| # | Severity | Location |
|---|----------|---------|
| 1 | **Blocking** | `useBalanceHistory`, `useDelegateDelegationHistory`, `useDelegationHistory` — duplicate `getNextPageParam` |
| 2 | **Blocking** | `DelegationHistoryTable` — silent removal of amount sort |
| 3 | **Blocking** | `useBalanceHistoryGraph`, `useDelegateDelegationHistoryGraph` — undocumented `limit: 1000` truncation |
| 4 | Major | `offchainVoteLabel.ts` — magic numbers + fragile comma-split multi-choice detection |
| 5 | Major | `ProposalsTable` / `DelegateProposalsActivity` — `string` prop type forcing `as` cast |
| 6 | Major | `useBalanceHistoryGraph` — `isLoading` vs `loading` naming inconsistency |
| 7–9 | Minor | `computedHasNextPage` alias, `StringifyBigInt` indirection, tooltip `max-w` duplication |


---

PR #1947, opened 2026-05-27
86ahp2bjc: migrate token distribution to Kubb SDK

pikonha:
## Summary

- Replace `useGetProposalsQuery` (GraphQL) with `useProposals` from `@anticapture/client/hooks` in the token-distribution feature hook
- Replace `useTokenMetricsLazyQuery` (GraphQL) with the `tokenMetrics` REST SDK function in the shared `useTimeSeriesData` hook
- Replace `useHistoricalTokenDataQuery` (GraphQL) with `useHistoricalTokenData` from `@anticapture/client/hooks` in `useDaoTokenHistoricalData`
- Update `useChartMetrics` to use the new REST response shape (`.items` instead of `.proposals.items`, `isLoading` instead of `loading`)

## Approach

Each GraphQL query is mapped to its REST equivalent from the Kubb-generated `@anticapture/client` package:

| Old (GraphQL) | New (REST SDK) |
|---|---|
| `useGetProposalsQuery` | `useProposals` from `@anticapture/client/hooks` |
| `useTokenMetricsLazyQuery` | `tokenMetrics` SDK function from `@anticapture/client` |
| `useHistoricalTokenDataQuery` | `useHistoricalTokenData` from `@anticapture/client/hooks` |

The DAO identifier moves from an Apollo context header (`anticapture-dao-id`) to a URL path parameter, which is the REST API pattern.

## Files changed and why

- **`features/token-distribution/hooks/useProposals.ts`** — the only file in token-distribution that directly imported graphql-client; replaced with the Kubb hook
- **`features/token-distribution/hooks/useChartMetrics.tsx`** — updated to use new return types: `isLoading` (React Query) instead of `loading` (Apollo), and `data.items` instead of `data.proposals.items`
- **`shared/hooks/useTimeSeriesData.ts`** — shared hook that powered the token-metrics timeseries; migrated to call the REST SDK function in a `Promise.all` preserving the existing manual state-machine pattern
- **`features/attack-profitability/hooks/useDaoTokenHistoricalData.ts`** — shared hook used by token-distribution for the TOKEN_PRICE metric; migrated to `useHistoricalTokenData`; returns `loading` (mapped from `isLoading`) to preserve backward compatibility with attack-profitability and panel callers

## Assumptions I made

- The REST API returns `TokenMetricItem.high` and `TokenMetricItem.volume` as strings at runtime (despite the Kubb type saying `bigint` due to the EthereumGenerator). The mapping code uses `String(item.high)` and `String(item.volume)` to safely convert to the local `TokenMetricItem` type.
- `useDaoTokenHistoricalData` is shared with attack-profitability and panel features; its external API (returning `{ data, loading, error }`) is preserved to avoid touching those callers.

## What I did NOT do

- Did not change callers of `useDaoTokenHistoricalData` in attack-profitability or panel (backward-compatible)
- Did not change the local `TokenMetricItem` type shape (kept as `string` to avoid cascading changes to mocked data and other consumers)
- Did not migrate `useTimeSeriesData` to use `useTokenMetrics` React Query hook — kept the manual `useCallback`/`useEffect` pattern to preserve the refresh interval and parallel-fetch behaviour

## Open questions for review

1. Should the local `TokenMetricItem` type (`high: string, volume: string`) be updated to match the Kubb SDK type (`high: bigint, volume: bigint`)? This would allow removing the `String()` mapping in `useTimeSeriesData`.
2. `useTimeSeriesData` might be a candidate for full React Query migration (`useQueries`) in a follow-up; this PR intentionally preserves the existing state machine.

## Link

ClickUp task: https://app.clickup.com/t/86ahp2bjc

---
_Generated by [Claude Code](https://claude.ai/code/session_01Wr71Qt1BvADmKo8ZycQDar)_


pikonha:
**Dashboard E2E — investigation findings**

The E2E check has failed twice (consistently, not a flake). I've traced what changed but cannot see the Playwright output to confirm the exact error. Here's what I found:

**What this PR changes at the network level:**
Before this migration, `useTimeSeriesData`, `useDaoTokenHistoricalData`, and the token-distribution `useProposals` fetched data via Apollo GraphQL → `NEXT_PUBLIC_BASE_URL` (external URL, never through `localhost`).

After this migration those same hooks make REST calls through the Next.js proxy → `http://localhost:3000/api/gateful/ens/token-metrics`, `…/ens/token/historical-data`.

**Why that matters for E2E:** The `watch5xxErrors` fixture in `e2e/helpers/network.ts` throws on any `localhost` response with status ≥ 500. Since GraphQL calls went to an external URL they were never monitored. The new REST calls DO go through localhost and will trigger this watcher if the upstream returns a 5xx.

**What to check:**
1. Download the Playwright report artifact from the failing CI run (Actions → `playwright-report` artifact) to get the actual error message — is it a `5xx response: 500 http://localhost:3000/api/gateful/ens/…` or a timeout/assertion failure?
2. If it's a 5xx: verify that the E2E Gateful environment (`E2E_NEXT_PUBLIC_GATEFUL_URL`) responds correctly to `GET /ens/token-metrics?metricType=DELEGATED_SUPPLY&…` and `GET /ens/token/historical-data`.
3. If it's a timeout/assertion: something in rendering is broken — the static text "TOKEN DISTRIBUTION SUPPLY" should always be visible regardless of API state, so a timeout there would indicate a React crash.

All other CI checks (typecheck, lint, changeset, codegen) pass.

---

apps/dashboard/features/attack-profitability/hooks/useDaoTokenHistoricalData.ts
```diff
@@ -25,27 +13,21 @@ export const useDaoTokenHistoricalData = ({
...
+  );
 
-  const items =
-    data?.historicalTokenData
-      ?.filter((item): item is HistoricalTokenDataItem => item !== null)
-      .map(toPriceEntry) ?? [];
+  const items = (data ?? []).map((item) => ({
+    price: item.price,
+    timestamp: item.timestamp,
+  }));
```
pikonha: remove this

apps/dashboard/features/token-distribution/hooks/useProposals.ts
```diff
@@ -1,19 +1,11 @@
...
-    },
-    variables: {
-      fromDate,
-      limit: 1000,
-    },
-    fetchPolicy: "cache-and-network",
-    notifyOnNetworkStatusChange: true,
+  return useProposalsSDK(daoId.toLowerCase() as ProposalsPathParamsDaoEnumKey, {
+    fromDate,
+    limit: 1000,
```
pikonha, line 9: add the lean query param so it doesn't return the description

PR #1971, opened 2026-06-11
feat: per-tenant API tokens - Authful service + Gateful auth (DEV-758)

brunod-e:
﻿## Summary

Replaces the single hardcoded `ANTICAPTURE_MCP_API_KEY` / `BLOCKFUL_API_TOKEN` shared secrets with **per-tenant API tokens** that carry rate limiting and usage tracking (billing-ready). Gateful becomes the single auth guardian for both MCP and direct REST traffic; a new **Authful** service owns token issuance, validation and revocation. Everything ships behind env flags — **merging this PR changes nothing in production** until the rollout envs are flipped.

## Context

- 📋 **Ticket**: [DEV-758 — Generate Gateful token on demand](https://app.clickup.com/t/86ahw07xd)
- 📐 **Design spec**: `docs/specs/dev-758-gateful-token-service.md`
- 🌿 **Branch**: `feat/gateful-api-tokens`

## The problem

Every external MCP consumer (including Uniswap) shares one static key with zero visibility of who calls what, and the MCP server swaps the caller's identity for an internal key upstream — so tenant identity never reaches Gateful. No rate limiting, no usage attribution, no revocation path.

### New service: `apps/authful` (+ `infra/authful/`)
- Hono + Drizzle + dedicated Postgres (schema `authful`): `tokens` + `usage_hourly` tables
- Admin surface (`ADMIN_API_KEY`): `POST /tokens` (mint — plaintext returned exactly once, only sha256 stored), `GET /tokens`, `DELETE /tokens/:id` (idempotent revoke)
- Internal surface (`INTERNAL_API_KEY`, called by Gateful): `POST /validate`, `POST /usage/batch` (additive upsert per token/route/hour)
- `pnpm authful mint` CLI — generates new tokens or **seeds an existing credential without rotating it** (`TOKEN_PLAINTEXT` env, never argv) — this is the Uniswap migration path
- Dockerfile runs `db:migrate` automatically before serving

### Gateful: `apps/gateful/src/auth/` (enabled only when `TOKEN_SERVICE_URL` is set)
- **`token-auth`**: bearer → sha256 → Redis cache (300s positive / 60s negative) → Authful on miss. Unknown/revoked → `401` fail-closed. Authful down **with** cached verdict → request proceeds; down **without** cache → `503` (distinguishable from bad credentials). Redis down → falls through to Authful (fail-open on cache errors). Plaintext never logged.
- **`rate-limit`**: fixed window per token/minute (limit stored per token), `429` + `Retry-After`. Fails open without Redis — availability over enforcement.
- **`usage`**: in-memory counters flushed to Authful every 30s, bounded-cardinality route normalization (`/ens/proposals/0x123` → `/{dao}/proposals`). Never blocks a request; failed flushes re-buffer.
- Without `TOKEN_SERVICE_URL`, the legacy `BLOCKFUL_API_TOKEN` path is untouched (refactored to a shared `PUBLIC_PATHS` set, behavior-identical — 77/77 existing tests pass).

### MCP server: `packages/anticapture-client` (enabled only when `FORWARD_CLIENT_AUTH=true`)
- Forwards each caller's own `Authorization` header to Gateful per-request via `AsyncLocalStorage` + axios interceptor, instead of swapping identities for the shared upstream key. Validated by a concurrency spike (two tenants, 10 interleaved calls, zero cross-contamination; spike removed after validation).

## Flow (new path)

```mermaid
sequenceDiagram
    participant C as Client (Uniswap / Blockful / 3rd party)
    participant M as MCP server
    participant G as Gateful
    participant R as Redis
    participant T as Authful (+Postgres)

    C->>M: tool call (Bearer tenant-token)
    Note over M: no validation — stores header in ALS
    M->>G: REST call (same Bearer, via axios interceptor)
    G->>R: GET token:sha256(token)
    alt cache miss
        G->>T: POST /validate {tokenHash}
        T-->>G: {valid, tenant, rateLimitPerMin}
        G->>R: SET (TTL 300s / 60s negative)
    end
    G->>R: INCR rl:tokenId:minute (429 if over limit)
    G->>G: proxy to DAO APIs
    G-->>M: response
    M-->>C: tool result
    Note over G,T: usage counters flushed in 30s batches
```

## Rollout (zero downtime, zero client action)

1. Deploy Authful + Postgres; **seed the current Uniswap MCP key and internal key as tenants** `uniswap` / `blockful` — same key values, so nothing breaks
2. Set `TOKEN_SERVICE_URL` + `TOKEN_SERVICE_API_KEY` on Gateful (dev first)
3. Set `FORWARD_CLIENT_AUTH=true` on the MCP server
4. After soak: remove `ANTICAPTURE_MCP_API_KEY` validation from MCP and the legacy Gateful branch (follow-up PR)

Rollback at any step = unset the env var. **Uniswap keeps using the exact same URL + header throughout.**

## Test plan

- `authful`: 13 integration tests (PGlite) — auth boundaries, mint/seed/409, revoke idempotency, validate + `last_used_at`, usage upsert + hour truncation
- `gateful`: 18 new tests covering the full resilience table (fail-closed 401, 503 only on outage-without-cache, cached-token survival through Authful outage, rate-limit fail-open, usage re-buffering, route normalization) — 77/77 total
- Manually verified locally end-to-end: 401 (no/wrong token) → 200 → 429 (rate limit) → revoke → 401 → usage rows in Postgres → MCP forward path

## Self-review summary

- ✅ typecheck / lint / build pass on all three packages; authful 13/13, gateful 77/77
- ⚠️ Known limitations (deferred to fast-follow, all in the best-effort usage path — none affect auth correctness):
  - Usage flush re-buffers on **any** error incl. 4xx; a poison entry could grow the buffer unbounded (should drop on 4xx, retry only network/5xx)
  - `normalizeRoute` passes unknown non-DAO first segments through verbatim (authed path probing could bloat `usage_hourly`; should bucket to `/unknown` like the cache middleware)
  - `AuthfulClient.validate` trusts the JSON response shape without schema validation
  - Revoked tokens stay valid up to the 300s cache TTL (by design — confirm acceptable)
  - Authful `/docs` is unauthenticated (fine if the Railway service is private-network only)
  - Gateful with `TOKEN_SERVICE_URL` but **no** `REDIS_URL` makes Authful a hard dependency for all traffic (no startup warning)

## Changesets

- `@anticapture/authful`: minor — new service
- `@anticapture/gateful`: minor — optional per-tenant token auth, rate limiting, usage tracking
- `@anticapture/client`: minor — per-request auth forwarding behind `FORWARD_CLIENT_AUTH`


packages/anticapture-client/src/configure-upstream-client.ts
```diff
@@ -11,4 +14,12 @@ export function configureUpstreamClient(): void {
       ...(apiKey ? { Authorization: `Bearer ${apiKey}` } : {}),
```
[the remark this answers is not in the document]
brunod-e: Fixed in 4a85cf47d: when `FORWARD_CLIENT_AUTH=true` the default `Authorization` header is no longer installed, so requests with no inbound auth reach Gateful anonymous and get a per-tenant 401 instead of riding the shared `ANTICAPTURE_API_KEY`.

pikonha (changes requested):
# Code Review

## Overview

Replaces the shared `BLOCKFUL_API_TOKEN`/`ANTICAPTURE_MCP_API_KEY` secrets with per-tenant tokens: a new **Authful** service (Hono + Drizzle, sha256-only storage) handles mint/validate/revoke/usage; **Gateful** gains opt-in middleware for token auth (Redis-cached), per-token rate limiting, and batched usage tracking; the **MCP server** can forward each caller's own `Authorization` header via `AsyncLocalStorage`. Everything is gated behind env flags, and the legacy path is refactored but behavior-identical.

## Strengths

- **Excellent failure-policy discipline.** The resilience matrix (fail-closed 401 for bad tokens, 503 only when Authful is down *and* no cached verdict, fail-open on Redis errors, fail-open rate limiting) is clearly reasoned, documented in the code, and each branch has a dedicated test. This is the hardest part of an auth gateway to get right and it's done well.
- **Secret hygiene is consistent:** plaintext never stored (sha256 only), never logged, never passed via argv (`TOKEN_PLAINTEXT` env in the mint script), `GET /tokens` returns metadata only, and the test asserts the stored row doesn't contain the plaintext.
- **Correct upsert handling:** `TokensService.recordUsage` pre-aggregates by primary key before `INSERT … ON CONFLICT` (which can't touch the same row twice), and Authful defensively re-truncates hours. Both covered by tests.
- Config validation via `.refine` (`TOKEN_SERVICE_API_KEY` required with `TOKEN_SERVICE_URL`), `unref()`'d flush timer, and the `PUBLIC_PATHS` refactor keeping the legacy branch identical are all nice touches.
- Changesets present and correctly scoped; CI's turbo `--filter=...[range]` picks up the new package automatically.

## Blocking: usage pipeline has a concrete poison-batch failure mode

The self-review lists three "deferred" limitations separately, but together they form one real bug:

- `UsageTracker.record` builds keys as `${tokenId}|${route}|${hour}` and `flush` re-splits on `|` (`apps/gateful/src/auth/usage.ts`).
- `normalizeRoute` passes unknown first segments through **verbatim** (`return \`/${first}\``).
- Authful's `UsageBatchBodySchema` zod-rejects the whole batch (400) on any invalid entry, and `flush` re-buffers on **any** error.

So one authenticated request to a path whose first segment contains a `|` (or anything that breaks the split/ISO-hour parse) produces an entry that makes Authful 400 the batch → flush throws → the *entire* batch re-buffers → every subsequent flush fails forever and the buffer grows unbounded. All usage tracking silently stops until restart — a silent-failure mode in the very feature this PR exists to ship.

**Each of these alone breaks the chain, all three are cheap:**
- Bucket unknown first segments to `/unknown` in `normalizeRoute` (the cache middleware already does this — mirror it).
- Use a structured buffer key (e.g. nested `Map` or `JSON.stringify` tuple) instead of `|`-joined strings.
- In `flush`, drop the batch on 4xx and only re-buffer on network/5xx errors.

## MCP rollout: per-tenant identity doesn't reach Gateful until step 4

`mcp-server-http.ts` still enforces `ANTICAPTURE_MCP_API_KEY` equality (`isValidToken`). With `FORWARD_CLIENT_AUTH=true` and that key still set, every MCP caller must present the *shared* key — so the forwarded header is always the same token, and all MCP traffic lands on **one tenant** in Authful. Per-tenant attribution via MCP only begins after the follow-up PR removes MCP-side validation. This is consistent with the rollout plan but easy to misread from the PR description ("forwards each caller's own Authorization"); worth stating explicitly in the rollout section so nobody expects per-tenant MCP metrics at step 3.

## Minor correctness/robustness items (non-blocking)

- **Rate limit `incr`/`expire` not atomic** (`rate-limit.ts`): if the process dies (or `expire` fails) between the two calls, the per-minute key persists in Redis without TTL forever. Harmless for correctness (keys are minute-scoped) but accumulates garbage. A Lua script or `SET … EX NX` + `INCR` pattern fixes it; fine as a follow-up.
- **429s are invisible to usage tracking**: `rateLimitMiddleware` short-circuits before `usageMiddleware`, so throttled requests are never counted. For billing/abuse analysis you likely *want* to see throttled volume. Decide intentionally.
- **No graceful-shutdown flush in Gateful**: the MCP server handles SIGTERM but `apps/gateful/src/index.ts` never calls `usageTracker.stop()`/`flush()`, so up to 30s of usage is lost on every deploy. One `process.on("SIGTERM")` away.
- **Mint script bypasses validation**: the HTTP surface enforces `plaintext.min(16)` via zod, but `scripts/mint-token.ts` passes `TOKEN_PLAINTEXT` straight to `TokensService.mint` with no length check. Move the min-length rule into the service so both paths share it.
- **Revoked hashes can never be re-seeded**: the `token_hash` unique constraint includes revoked rows, so accidentally revoking the seeded Uniswap key cannot be undone by re-seeding (409) — you'd be forced to rotate. Probably acceptable, but worth knowing during the migration window.
- **Unauthenticated negative-verdict caching**: any client can spam random bearers, each costing one Authful round-trip plus one 60s Redis key. Bounded by TTL, but Authful has no protection of its own; keep it private-network-only (as planned) and consider a max on cached negatives later.

## Conventions

- The unvalidated `as TokenValidation` cast in `authful-client.ts` is acknowledged in the self-review but is cheap to fix now with a small zod schema, and it guards your own 503-vs-401 logic against Authful contract drift. (The `db as any` in the integration test is a pragmatic drizzle-kit `pushSchema` limitation, explicitly lint-disabled — acceptable.)
- Two changesets both bump `@anticapture/client` minor — harmless (same resulting version, two changelog lines), but the `dev-758-authful-gateful.md` entry already covers the client change; you could drop `dev-758-mcp-forward-auth.md`.
- The PR carries a **Dashboard** label but touches no dashboard code.

## Test coverage

Strong overall: 13 PGlite integration tests on Authful (auth boundaries between admin/internal keys, seed/409, idempotent revoke, hour truncation, bigint accumulation) and 18 Gateful unit tests covering every row of the failure-policy table. Gaps worth adding: a `normalizeRoute` case for a hostile/unknown segment (would have surfaced the poison-batch issue), and a `flush` test where the client rejects with a 4xx-style error.

## Verdict

High-quality, well-tested, and the env-flag rollout genuinely makes this a no-op to merge. Requesting changes only for the **usage poison-batch chain** — please also add the `TokenValidation` response schema while in there. Everything else can be the stated fast-follow.

---

apps/api/src/controllers/draft-proposals/index.ts
```diff
@@ -21,7 +21,7 @@ export function draftProposals(
     createRoute({
       method: "get",
       operationId: "getDraftProposals",
-      path: "/proposal/drafts",
+      path: "/proposal-drafts",
```
LeonardoVieira1630, line 24: why? It is not restful anymore.

apps/authful/scripts/mint-token.ts
LeonardoVieira1630, line 1: Is this script necessary? I mean, the mint is being done by http request right? So why maintain this?

apps/authful/src/controllers/usage/index.ts
LeonardoVieira1630, line 1: Remove this endpoint. Usage is something that is dealt on gateful and observability needs to be done on prometheus. We dont need this data being pushed on this Service (authful).

apps/authful/src/database/schema.ts
```diff
@@ -0,0 +1,39 @@
...
+    route: d.text().notNull(), // normalized route pattern, e.g. /{dao}/proposals
+    hour: d.timestamp({ withTimezone: true }).notNull(), // truncated to the hour
+    // No default: count always comes from the batch payload, and a bigint
+    // default breaks drizzle-kit's snapshot serialization (JSON + BigInt).
+    count: d.bigint({ mode: "bigint" }).notNull(),
+  }),
+  (table) => [
+    primaryKey({ columns: [table.tokenId, table.route, table.hour] }),
+  ],
+);
```
LeonardoVieira1630: Remove this table. Observability is done on prometheus.

apps/authful/src/middlewares/logger.ts
```diff
@@ -0,0 +1,34 @@
...
+      logger.error(
+        {
+          err,
+          method: c.req.method,
+          path: c.req.path,
+          durationMs: Math.round(performance.now() - start),
+        },
+        "request failed",
+      );
+      throw err;
```
LeonardoVieira1630, line 21: Using a catch just to throw after it? Is it worth?

docs/specs/dev-758-gateful-token-service.md
LeonardoVieira1630, line 1: remove this file

apps/authful/src/repositories/tokens/index.ts
```diff
@@ -0,0 +1,63 @@
...
+    if (entries.length === 0) return;
+    await this.db
+      .insert(usageHourly)
+      .values(entries)
+      .onConflictDoUpdate({
+        target: [usageHourly.tokenId, usageHourly.route, usageHourly.hour],
+        set: {
+          count: sql`${usageHourly.count} + excluded.count`,
+        },
+      });
```
LeonardoVieira1630: remove this

apps/authful/src/env.ts
```diff
@@ -0,0 +1,15 @@
...
+
+dotenv.config();
+
+const envSchema = z.object({
+  PORT: z.coerce.number().default(4002),
+  DATABASE_URL: z.string(),
+  // Guards human-facing token management endpoints (mint/list/revoke).
+  ADMIN_API_KEY: z.string().min(16),
+  // Guards service-facing endpoints (/validate, /usage/batch), shared with Gateful.
+  INTERNAL_API_KEY: z.string().min(16),
```
LeonardoVieira1630, line 12: do we need too keys here?

apps/authful/src/services/tokens/index.ts
```diff
@@ -0,0 +1,93 @@
...
+   * new one. Used to migrate pre-existing shared keys (e.g. Uniswap's current
+   * MCP key) into the store without rotating them.
+   */
+  plaintext?: string;
+};
+
+export type MintedToken = {
+  token: DBToken;
+  /** Returned exactly once at mint time; only the hash is persisted. */
+  plaintext: string;
```
LeonardoVieira1630, line 20: We should not have this. 
- With this, an API-key would travel in the body of a request. This is a security issue. 
- This exists just because of ONE single pre existing key. Instead of having a production code coupled with the migration problem, why dont we manually insert the old row on the dB?

apps/authful/src/services/tokens/index.ts
```diff
@@ -0,0 +1,93 @@
...
+    // one batch must be merged before hitting the database.
+    const merged = new Map<string, UsageEntry>();
+    for (const entry of entries) {
+      const key = `${entry.tokenId}|${entry.route}|${entry.hour.getTime()}`;
+      const existing = merged.get(key);
+      if (existing) existing.count += entry.count;
+      else merged.set(key, { ...entry });
+    }
+    await this.repo.recordUsage([...merged.values()]);
+  }
```
LeonardoVieira1630: remove this

apps/authful/src/metrics.ts
LeonardoVieira1630, line 1: Is this really util / necessary?

apps/authful/drizzle.config.ts
```diff
@@ -0,0 +1,14 @@
...
+import { defineConfig } from "drizzle-kit";
+
+dotenv.config();
+
+export default defineConfig({
+  schema: "./src/database/schema.ts",
+  out: "./drizzle",
+  dialect: "postgresql",
+  dbCredentials: {
+    url: process.env.DATABASE_URL!,
```
LeonardoVieira1630, line 11: dont we use zod? Why are we just relaying on process here?

apps/gateful/src/auth/authful-client.test.ts
LeonardoVieira1630, line 1: is this test useful?

apps/gateful/src/auth/rate-limit.test.ts
```diff
@@ -0,0 +1,88 @@
...
+  });
+
+  it("fails open without a store", async () => {
+    const res = await buildApp(undefined).request("/test");
+    expect(res.status).toBe(200);
+  });
+
+  it("fails open when the store errors", async () => {
+    const broken: RateLimitStore = {
+      incr: () => Promise.reject(new Error("redis down")),
```
LeonardoVieira1630, line 82: redis?

apps/gateful/src/auth/rate-limit.test.ts
LeonardoVieira1630, line 1: Is this test useful / necessary? I mocks a lot of things.

apps/gateful/src/auth/rate-limit.ts
```diff
@@ -0,0 +1,46 @@
...
+    const epochMinute = Math.floor(Date.now() / 60_000);
+    const key = `rl:${auth.tokenId}:${epochMinute}`;
+
+    let count: number;
+    try {
+      count = await store.incr(key);
+      // First hit of the window sets the expiry; 2 minutes covers clock skew.
+      if (count === 1) await store.expire(key, 120);
+    } catch (err) {
+      logger.warn({ err }, "rate limit store failed — failing open");
```
LeonardoVieira1630, line 33: Do we want a fail open here? On relayer i did a fail close.

apps/gateful/src/auth/token-auth.test.ts
```diff
@@ -0,0 +1,163 @@
...
+
+const VALID: TokenValidation = {
+  valid: true,
+  tokenId: "11111111-1111-1111-1111-111111111111",
+  tenant: "uniswap",
+  rateLimitPerMin: 600,
+};
+
+function fakeClient(impl: () => Promise<TokenValidation>) {
+  return { validate: vi.fn(impl) } as unknown as AuthfulClient & {
```
LeonardoVieira1630: do we really need as unknown as AuthfulClient?

apps/gateful/src/auth/token-auth.test.ts
```diff
@@ -0,0 +1,163 @@
...
+}
+
+const VALID: TokenValidation = {
+  valid: true,
+  tokenId: "11111111-1111-1111-1111-111111111111",
+  tenant: "uniswap",
+  rateLimitPerMin: 600,
+};
+
+function fakeClient(impl: () => Promise<TokenValidation>) {
```
LeonardoVieira1630: do we need a fake client? Could we use MSW?

apps/gateful/src/auth/token-auth.ts
```diff
@@ -0,0 +1,114 @@
...
+import { TokenValidationSchema } from "./authful-client.js";
+import type { AuthfulClient, TokenValidation } from "./authful-client.js";
+
+export type AuthContext = {
+  tokenId: string;
+  tenant: string;
+  rateLimitPerMin: number;
+};
+
+declare module "hono" {
```
LeonardoVieira1630, line 21: why?

packages/anticapture-client/src/configure-upstream-client.ts
```diff
@@ -1,14 +1,30 @@
-import { setConfig } from "@kubb/plugin-client/clients/axios";
+import { axiosInstance, setConfig } from "@kubb/plugin-client/clients/axios";
+
+import { inboundAuthStorage } from "./request-context.ts";
 
 export function configureUpstreamClient(): void {
   const baseURL = process.env["ANTICAPTURE_API_URL"] ?? "http://localhost:4001";
   const apiKey = process.env["ANTICAPTURE_API_KEY"];
```
LeonardoVieira1630: zod?

packages/anticapture-client/mcp-server-http.ts
```diff
@@ -1,11 +1,14 @@
...
 import { StreamableHTTPServerTransport } from "@modelcontextprotocol/sdk/server/streamableHttp";
 import { createMcpServer } from "./src/create-mcp-server.ts";
 import { configureUpstreamClient } from "./src/configure-upstream-client.ts";
+import { inboundAuthStorage } from "./src/request-context.ts";
+import { AuthfulClient, hashBearerToken } from "./src/authful-client.ts";
 import pino from "pino";
 
-const mcpApiKey = process.env["ANTICAPTURE_MCP_API_KEY"];
+const tokenServiceUrl = process.env["TOKEN_SERVICE_URL"];
+const tokenServiceApiKey = process.env["TOKEN_SERVICE_API_KEY"];
```
LeonardoVieira1630: zod?

apps/authful/src/database/schema.ts
PedroBinotto: this is for billing, not observability

apps/authful/src/controllers/usage/index.ts
PedroBinotto: this is for billing, not observability

apps/authful/scripts/mint-token.ts
PedroBinotto: it's just a shorthand way of doing it (instead of `curl`ing it manually). constructing requests is easier and less error-prone this way. I think it's worth keeping

apps/api/src/controllers/draft-proposals/index.ts
PedroBinotto: this was a residual change from some other past PR that was not done properly and was causing stuff to break. if we do decide to change it to `/proposals/draft` (which I am all for btw) it should be done properly in a separate PR

apps/authful/src/middlewares/logger.ts
PedroBinotto: in this case, I think it is worth it as we get a log entry out of it

apps/authful/src/repositories/tokens/index.ts
PedroBinotto: as per the other comments, from what I gathered (since neither @pikonha nor @brunod-e are around to confirm) these metrics are for billing, not observability. I did add the standard monitoring stack to this PR separately

apps/authful/src/services/tokens/index.ts
PedroBinotto: I'm not sure I get what you mean. The client token must be returned to them regardless of the pre-existing key issue (which was solved exactly as you described). every new token that gets minted is returned exactly once so that the user may save and use it to identify themselves

apps/authful/src/services/tokens/index.ts
PedroBinotto: this is not observability, see other comments above

apps/authful/src/metrics.ts
PedroBinotto: yep. this is for exporting metrics to prometheus

apps/authful/drizzle.config.ts
PedroBinotto: I think you mean `dotenv`, no? if that's so, I agree that we should add it

apps/authful/src/env.ts
PedroBinotto: yep. the admin key is for "manual" (internal operator) tasks such as minting and revoking tokens. the other one is for requests between services (Gateful/MCP server -> Authful)

apps/api/src/controllers/draft-proposals/index.ts
LeonardoVieira1630: hmmm ok.

apps/gateful/src/auth/rate-limit.test.ts
PedroBinotto: not sure (this is @brunod-e's code), but test coverage doens't ever really hurt imo

apps/gateful/src/auth/authful-client.test.ts
PedroBinotto: not sure (this is @brunod-e's code), but test coverage doens't ever really hurt imo

apps/gateful/src/auth/rate-limit.test.ts
PedroBinotto: token interactions were implemented "cache-first"

apps/authful/scripts/mint-token.ts
LeonardoVieira1630: Really? I guess bruno said that he would remove that too. Like, maintain a file that cost one command to run, just to avoid a curl that is also one command....

apps/authful/src/controllers/usage/index.ts
LeonardoVieira1630: Billing? Whaaaat.

apps/gateful/src/auth/token-auth.test.ts
LeonardoVieira1630: use msw

apps/gateful/src/auth/token-auth.ts
PedroBinotto: It's exactly what Hono's docs recommend for typing context variables. The alternative — generics like new Hono<{ Variables: { auth: AuthContext } }>() — has to be repeated on every Hono instance, sub-router, and middleware
  signature, and they have to agree. Global augmentation types it once.

apps/gateful/src/auth/rate-limit.ts
PedroBinotto: this is just cache related. best to leave as fail-open

apps/authful/src/database/schema.ts
```diff
@@ -0,0 +1,21 @@
+import { index, pgSchema } from "drizzle-orm/pg-core";
+
+export const authfulSchema = pgSchema("authful");
+
+export const tokens = authfulSchema.table(
+  "tokens",
+  (d) => ({
+    id: d.uuid().primaryKey().defaultRandom(),
+    tenant: d.text().notNull(), // "uniswap", "blockful", ...
```
LeonardoVieira1630, line 9: Do we really need this and "name"?

LeonardoVieira1630 (approved):

apps/authful/drizzle/0000_oval_mattie_franklin.sql
```diff
@@ -0,0 +1,15 @@
...
+	"name" text NOT NULL,
+	"token_hash" text NOT NULL,
+	"rate_limit_per_min" integer DEFAULT 600 NOT NULL,
+	"created_at" timestamp with time zone DEFAULT now() NOT NULL,
+	"revoked_at" timestamp with time zone,
+	"last_used_at" timestamp with time zone,
+	CONSTRAINT "tokens_token_hash_unique" UNIQUE("token_hash")
+);
+--> statement-breakpoint
+CREATE INDEX "tokens_tenant_index" ON "authful"."tokens" USING btree ("tenant");
```
pikonha:
```suggestion
CREATE SCHEMA IF NOT EXISTS "authful";
--> statement-breakpoint
CREATE TABLE IF NOT EXISTS "authful"."tokens" (
	"id" uuid PRIMARY KEY DEFAULT gen_random_uuid() NOT NULL,
	"tenant" text NOT NULL,
	"name" text NOT NULL,
	"token_hash" text NOT NULL,
	"rate_limit_per_min" integer DEFAULT 600 NOT NULL,
	"created_at" timestamp with time zone DEFAULT now() NOT NULL,
	"revoked_at" timestamp with time zone,
	"last_used_at" timestamp with time zone,
	CONSTRAINT "tokens_token_hash_unique" UNIQUE("token_hash")
);
--> statement-breakpoint
CREATE INDEX IF NOT EXISTS "tokens_tenant_index" ON "authful"."tokens" USING btree ("tenant");
```

pikonha (approved):

PR #1996, opened 2026-06-24
fix: add auth to prometheus metrics scraping

PedroBinotto:


apps/gateful/src/index.ts
```diff
@@ -51,11 +52,24 @@ app.use("*", cors({ origin: "*" }));
 app.use("*", requestLogger());
 app.use("*", metricsMiddleware());
 
+// Protect the public /metrics endpoint with a shared bearer so only our
+// Prometheus scraper can read it. Registered before the route handler so the
+// guard runs first; skipped entirely when GATEFUL_METRICS_TOKEN is unset (local dev).
+if (config.metricsToken) {
+  app.use("/metrics", bearerAuth({ token: config.metricsToken }));
+}
```
PedroBinotto, line 60: auth for this route was done directly in the gateful instance (as opposed to doing it via authful) as infra requests (prometheus) don't seem to belong in the tenant/consumer model that we track via the auth service; this way it remains unpolluted and plain simpler

pikonha (approved):

PR #1997, opened 2026-06-25
feat(dashboard): create-proposal fixes — calldata builder, duplicate action, transfer UX

brunod-e:
## Summary

[DEV-916](https://app.clickup.com/t/86aj5nqdv) — a batch of fixes and improvements to the ENS **create-proposal** flow (`apps/dashboard/features/create-proposal`).

The headline change is a **recursive calldata builder** that handles every Solidity argument type, backed by an `argTree` value model that serializes to the existing `args[]` storage format — so **encoding is byte-identical and saved drafts stay compatible** (no migration).

## What's included

**Custom action — calldata builder**
- Recursive `ArgInput` renderer covering `bool`, `uintN`/`intN`, `address` (ENS), `bytesN`, `bytes`/`string`, dynamic arrays `T[]` (add/remove, empty allowed), fixed arrays `T[k]`, multidimensional arrays, and `tuple`/struct — nesting recursively (array of structs, struct of arrays, …).
- **Two-way paste & decode**: paste a `0x…` calldata blob → decodes against the function ABI into editable fields; a live **encoded-calldata preview** updates as you edit.
- Debounced contract-address validation (no error while still typing).

**Transfer action**
- Treasury (timelock) balance shown as a clickable **"Max"** (the contract that actually pays on execution).
- Helper text under Amount always visible; **per-token USD** for any ERC-20 with a CoinGecko price (`useTokenPrice`), neutral fallback otherwise.
- Clearer **selected** state on suggested-token badges.

**Actions list**
- **Duplicate action** button alongside edit/delete (independent deep copy).

**Dev surface**
- Storybook story (`Create Proposal/ArgInput`) exercising every supported type.

## Deferred (not in this PR)

- **Tenderly "Simulate" slice** is intentionally left out pending a plan decision (Tenderly's Simulation API requires a paid plan; the org is currently on Free). The work is staged locally and will land in a follow-up once that's resolved.

## Testing

- `pnpm --filter @anticapture/dashboard test` → **all suites pass** (added: `argTree`, `cloneAction`; plus existing `encodeActions`/`schema`/`fetchAbi` unchanged). New tests cover round-trips for every type incl. nesting, **no-regression equivalence vs the existing encoder**, and paste→decode→re-encode round-trips.
- `pnpm --filter @anticapture/dashboard typecheck` → clean.
- `pnpm --filter @anticapture/dashboard lint` → 0 errors.

## Reviewer notes

- Touches the same feature folder as #1990 (preview-drafts); branched off latest `dev` to minimize conflicts.
- No new runtime dependencies. USD now resolves uniformly via CoinGecko-by-contract (replaces the gov-token-only price path).


isadorable-png (commented):
ENS is not converting to USD
image

Can't read encoded calldata text, should be white.
But it's not needed to create a calldata field at the end, it could be placed in the top input, but instead of the button being "decode" it could be "clear" once it has data.
image

isadorable-png (approved):

PR #2002, opened 2026-06-28
feat: integrate tornado cash

pikonha:

PR #2006, opened 2026-06-29
fix(torn): correct quorum calc + governance-config facts (review follow-up)

Zeugh-eth:
## What

Small, **factually-verifiable** corrections to the TORN integration, split out from my review on #2002 so they're easy to review/merge. Targets `feat/tornado-cash-integration` to fold into #2002.

### Included (low-risk, verified against the on-chain contract)
- **`apps/api/src/clients/torn/index.ts` — `calculateQuorum`**: count `forVotes + againstVotes`. `Governance.state()` reaches quorum on `forVotes + againstVotes >= QUORUM_VOTES` (both sides count), so returning `forVotes` alone could mislabel `NO_QUORUM`. Status docstring updated to match.
- **`apps/dashboard/shared/dao-config/torn.ts`**:
  - Proposal threshold copy: **1,000 TORN** (was 25,000 — that's the old quorum figure; current `PROPOSAL_THRESHOLD = 1e21`).
  - Voting delay copy: **75 seconds** (was "~1 block (~12 seconds)"; on-chain `VOTING_DELAY = 75`). Risk rating left HIGH — 75s is still very short.
  - `rules.changeVote: true` — re-voting is allowed on-chain (`_castVote` overwrites the prior receipt while a proposal is Active).

### Deliberately NOT included (need your call / testing)
- **Re-vote tally handling** in `apps/indexer/src/indexer/torn/governor.ts`: the `onConflictDoNothing` on `(voter, proposalId)` drops genuine re-votes (not just backfill replays), so tallies can drift from chain. A correct fix must stay idempotent on replay (e.g. compare stored `txHash`/`logIndex`) — left to you since it needs an indexer run to verify.
- **`VOTE_MUTABILITY` risk field** still reads "does not allow changing votes"; with `changeVote` now `true`, that field's copy/rating likely wants updating too — left as a product/risk-scoring call.
- **`rules.logic: "For"`** is now inconsistent with the quorum fix; "All Votes Cast" (for+against, abstain always 0) would match — didn't change it to avoid touching the enum without your confirmation.

> No behavioral logic changed in the indexer; CI will typecheck. Happy to fold in the indexer re-vote fix if you want it here.

### DIFF
--- a/apps/api/src/clients/torn/index.ts
+++ b/apps/api/src/clients/torn/index.ts
@@ -105,7 +105,9 @@ export class TORNClient<
     againstVotes: bigint;
     abstainVotes: bigint;
   }): bigint {
-    return votes.forVotes;
+    // Tornado's Governance.state() reaches quorum on forVotes + againstVotes
+    // (both sides count toward QUORUM_VOTES), not forVotes alone.
+    return votes.forVotes + votes.againstVotes;
   }
 
   alreadySupportCalldataReview(): boolean {
@@ -123,7 +125,7 @@ export class TORNClient<
    *   EXECUTED → finalized (persisted by indexer)
    *   now < startTime → PENDING
    *   now < endTimestamp → ACTIVE
-   *   forVotes < quorum → NO_QUORUM
+   *   forVotes + againstVotes < quorum → NO_QUORUM
    *   forVotes <= againstVotes → DEFEATED
    *   now < endTimestamp + EXECUTION_DELAY → QUEUED
    *   now < endTimestamp + EXECUTION_DELAY + EXECUTION_EXPIRATION → PENDING_EXECUTION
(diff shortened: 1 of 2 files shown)


Zeugh-eth:
Both addressed in `4e60b18e` (thanks @codex):
- **quorum label** — `rules.logic` → `"All Votes Cast"` so the dashboard (MetricsCard / governance-settings) matches `calculateQuorum` now counting for+against.
- **vote mutability** — `VOTE_MUTABILITY` now **LOW** with copy reflecting that TORN allows changing votes (consistent with `changeVote: true`); also fixed the `attackExposure` line that claimed 'immutable votes'.

This makes the quickfix PR internally consistent. The deeper items (indexer re-vote tally, Vault custody, per-account voting power, target-verification detector) are tracked for the full implementation.

Zeugh-eth:
@codex review

pikonha (approved):

PR #2013, opened 2026-06-30
fix(dashboard): encrypt vote choice offchain proposals

brunod-e:
## Problem

Voting on offchain proposals with **Shutter** privacy (e.g. ENS Copeland working-group elections) failed with:

```json
{"error":"client_error","error_description":"invalid choice"}
```

Shutter proposals require the vote `choice` to be **Shutter-encrypted** (a `0x…` string); the dashboard was sending the plaintext ranked array (e.g. `[5,1,2,7,3,4,6]`). The Snapshot sequencer rejects any non-encrypted choice on a shutter proposal.

## Root cause

Snapshot stores `privacy` on the proposal, not in the vote envelope, and we don't index it anywhere (it isn't even requested in the indexer's `PROPOSALS_QUERY`, so it never reaches the DB → API → client model). The dashboard therefore treated every offchain proposal the same and never encrypted.

## Fix

- Read `privacy` from the Snapshot hub right before voting; when it's `shutter`, encrypt the choice (mirroring Snapshot's reference `encryptChoice` byte-for-byte) and submit with `privacy: "shutter"` so snapshot.js signs it as a string vote.
- New `encryptChoice` util with memoized WASM init (lazy import keeps the 330KB module out of the initial bundle).
- Added `@shutter-network/shutter-crypto@1.0.1` and vendored its `shutter-crypto.wasm` into `public/` (served as a static asset — robust across turbo dev / webpack build, unlike a pre-script copy).
- Declared `@ethersproject/{bignumber,bytes,random,strings}@5.8.0` explicitly (already transitively present; required by pnpm's strict resolution). These match Snapshot's reference exactly to guarantee identical byte output.

Why a live hub fetch instead of plumbing `privacy` through indexer → API → client: that's the structurally cleaner fix but spans 6 packages + codegen + contract changesets — out of scope for a hotfix. The live fetch's worst case (hub unreachable on a shutter proposal) is identical to today's behavior — a clear "invalid choice" rejection, never silent data corruption. Follow-up tracked to surface `privacy` through the data model.

## Verification

- **Real crypto, end-to-end:** encrypting the exact failing choice with the real EON public key + real proposal id produces a `0x` blob of **192 bytes — identical in length to actual successful votes** on this proposal (which only the sequencer accepts).
- Unit tests for the encoding contract (3/3 pass); full governance suite green.
- `typecheck` and `lint` clean.
- ⚠️ Still pending before merge: browser smoke test (confirm `/shutter-crypto.wasm` is served as `application/wasm`) and one real vote on an active shutter proposal.


isadorable-png:
## 🎨 UI Review

> Automated review · scope: `apps/dashboard/features/governance/components/modals/OffchainVotingModal.tsx` (all other changed files are non-UI: hooks, crypto util, types, deps, lockfile, changeset)
> Caveat: ⚠️ No Figma reference found (checked PR body/comments, both preconfigured ClickUp lists, and a workspace-wide ClickUp search) and no live dashboard preview was deployed for this PR (Vercel shows the `anticapture` project deployment as "Ignored" — only the Storybook project built, and it has no story for this component). Reviewed from diff + surrounding code context only.

---

### `OffchainVotingModal` — comment field / vote-gating for Shutter proposals

- **Layout shift on shutter proposals** — `isShutter` derives from `useOffchainProposalPrivacy`'s `data`, which is `undefined` while the hub fetch is in flight, so `isShutter` reads `false` during that window regardless of the real answer. The writable "Comment (optional)" textarea therefore renders first and gets swapped for the "encrypted, comments disabled" notice once the fetch resolves — a visible flip on exactly the proposals this PR is fixing (ENS Copeland elections). The file already has an established pattern for this: the voting-power block just above branches on `isVpLoading` to show an explicit "Loading..." placeholder instead of guessing at content. Suggest doing the same here — render a neutral loading placeholder for the comment slot while `isPrivacyLoading` is true, then swap to either the encrypted notice or the textarea once resolved. (`apps/dashboard/features/governance/components/modals/OffchainVotingModal.tsx:69-72`, `:308-334`) [UX-heuristic]

- **Vote button disables silently during the privacy check** — `isVoteEnabled` now also requires `!isPrivacyLoading` (`OffchainVotingModal.tsx:113-119`), so a user who already picked a choice and connected a wallet sees the Vote button go inert for the hub round-trip with no on-screen reason. The voting-power section right above handles the analogous case explicitly (`isVpLoading` → "Loading...", `vpError` → "Unable to fetch voting power"). Consider a matching one-line hint (e.g. "Checking proposal privacy…") so this doesn't read as a stuck/broken button — worth being deliberate here, since a confusing disabled state is exactly the kind of thing that makes the original "invalid choice" failure hard to diagnose from a user's side. [UX-heuristic]

---

No other UI-visible surface in this diff — `useOffchainProposalPrivacy.ts` and `useVoteOnOffchainProposal.ts` are logic-only hooks, `shutter.ts` / `shutter.test.ts` are crypto utils (out of scope for a UI review), and `package.json` / `pnpm-lock.yaml` / `types/shutter-crypto.d.ts` / `public/shutter-crypto.wasm` are dependency wiring.

---

pikonha (approved):

PR #2036, opened 2026-07-08
Dev

pikonha:


isadorable-png:
## 🎨 UI Review

> Automated review · scope check: 82 files / 53 commits in this dev→main rollup, but ~19 touch dashboard-rendered UI (new error-boundary trio, offchain voting modal, proposal-overview components, whitelabel shell/sidebar, `ShutterIcon`, `theme.ts`). The remaining ~63 files (api/authful/gateful, infra/monitoring config, lockfile, CHANGELOGs, non-UI tests) are out of scope for this review.
> ⚠️ No Figma reference found (PR body has no figma.com/app.clickup.com link; ClickUp search for this PR/branch turned up nothing — expected, since "dev"/"Dev" are too generic to match). Vercel preview for this PR's `dev` environment shows as "Ignored" in the bot comment (no fresh build tied to this PR), and a direct guess at `anticapture-git-dev-ful.vercel.app` returned 403. Grounded in **diff + UX-expert evaluation only**, no visual confirmation.

---

### Error boundaries (`error.tsx` × 3 → `RouteErrorFallback`)

**Copy bug**
- The heading text has inconsistent source casing: `[ERROR:something_WENT_WRONG]` — `something` is lowercase while `WENT_WRONG` is upper-snake-case. The CSS `uppercase` class masks this visually, but the raw string ships as-is, so a user who selects/copies the error text (e.g. to paste into a bug report or Slack) gets `[ERROR:something_WENT_WRONG]` instead of consistent casing. Fix the source string itself rather than relying on `uppercase` to paper over it. (`apps/dashboard/shared/components/errors/RouteErrorFallback.tsx:31`) [Code-only]

**Consistency with other error states in this diff**
- Every other new/touched error state in this PR (`FetchErrorState.tsx:22`, and its usages in `ActivityFeedSection`, `OffchainVotesContent`, `TabsVotedContent`) renders through `BlankSlate` with an icon (`AlertOctagon`). `RouteErrorFallback` hand-rolls its own `div`/`h3`/`p` layout with no icon at all (`apps/dashboard/shared/components/errors/RouteErrorFallback.tsx:26-32`). For the full-page (highest-visibility) error surface to be the one that's least visually reinforced is backwards — worth at least adding an icon above the heading for hierarchy/consistency with the inline states. [UX-heuristic]

---

### `OffchainVotingModal.tsx`

**Missing dialog semantics**
- The modal container has no `role="dialog"`, `aria-modal="true"`, or `aria-labelledby` (`apps/dashboard/features/governance/components/modals/OffchainVotingModal.tsx:246`), and there's no focus trap or initial-focus management when it opens. Keyboard/screen-reader users can tab into content behind the backdrop, and assistive tech won't announce it as a dialog. Worth wrapping in the same dialog primitive/pattern as other modals in the app, if one exists, or adding the missing ARIA attributes + focus trap directly. [Code-only]

---

### `WhitelabelShell.tsx` / `WhitelabelSidebar.tsx`

**Duplicate nav config = mobile/desktop drift risk**
- Both files define their own copy of the exact same `NAV_ITEMS` array (`apps/dashboard/widgets/WhitelabelShell.tsx:27` for the mobile drawer, `apps/dashboard/widgets/WhitelabelSidebar.tsx:25` for the desktop sidebar). They currently match, but the next person who adds/reorders/re-icons a nav entry only has to touch one of the two to silently break mobile/desktop parity in the whitelabel shell — the component the skill flags as shared across every DAO theme. Worth extracting to one shared source of truth. [Code-only]

---

### `theme.ts`

**Link-contrast threshold looks low**
- `MIN_LINK_CONTRAST = 2` (line 62) gates whether a near-black brand color gets lightened for text/links. A contrast ratio of 2:1 is below WCAG's 3:1 minimum for non-text UI, so a brand color could clear this check and still read as barely-distinguishable from body text for low-vision users. Worth double-checking this threshold against an actual contrast target (3:1 minimum, ideally closer to 4.5:1 for text) rather than an arbitrary "distinguishable" cutoff. [Code-only]

---

- `FetchErrorState` (BlankSlate + `AlertOctagon` + inline "Try again" link) is a clean, consistent reuse of the existing empty-state pattern already used for filtered-empty states (e.g. "Clear filters" in `ActivityFeedSection`) — good pattern reuse, not a hand-rolled one-off.
- `RouteErrorFallback`'s body copy ("This page hit an unexpected error. The rest of the dashboard is unaffected.") is appropriately reassuring and free of stack traces/dev-jargon — good copy quality for a user-facing error boundary.
- `OffchainVotingModal`'s Shutter-proposal handling (hiding the comment field with an explanation instead of just disabling it silently) is a good example of explaining *why* a control is unavailable rather than leaving users guessing.
- `ForceErrorTrigger` is inert in production (`NODE_ENV` guard) — no user-facing risk from shipping a test hook in the page tree.


---

PR #1979, opened 2026-06-17
feat(dashboard): Shutter whitelabel governance frontend with Azorius proposal creation

Zeugh-eth:
## Summary

Brings the whitelabel governance frontend (already live for ENS) to **Shutter DAO**, with Shutter branding and full support for Shutter's **Azorius (Fractal)** governance stack — which differs from ENS's OZ Governor.

## What changed

**Enable + brand the whitelabel route (`shared/dao-config/shu.ts`)**
- Add `whitelabel: {}`, `hostnames: ["shutter.gov.blockful.io"]`, and `govPlatform`.
- Set the brand color to Shutter navy `#051016` (the whole light theme derives from `color.svgColor`).
- Enable the `activityFeed` page (Shutter feed events are indexed). Sidebar shows: Proposals, Holders & Delegates, Activity Feed, Notifications, Governance Settings.

**Branding (`shared/components/icons/ShutterIcon.tsx`)**
- Icon background `#1E1E1E` → `#051016` to match the brand.

**Proposal creation for Azorius (`features/create-proposal/…`)**
- `constants.ts`: allow `SHU` in `canCreateProposalForDao`.
- `usePublishProposal.ts`: same form/UI, branch the contract call by DAO id (consistent with the existing vote/execute switches). Shutter maps the encoded actions onto Azorius `Transaction[]` tuples + JSON `{title, description}` metadata and calls `submitProposal(strategy, 0x, transactions, metadata)` on the Azorius module; reads the new id from the Azorius `ProposalCreated` event. The OZ Governor path is unchanged.

## Governance mechanism verification (Shutter = Azorius, not OZ)
- **Vote** — already correct: `LinearERC20Voting.vote(uint32, uint8)` (`for=1, against=0, abstain=2`).
- **Delegate** — already correct: ERC20Votes `delegate(address)` on the SHU token.
- **Execute** — already correct: Azorius `executeProposal(...)`.
- **Propose** — was OZ-only/blocked; implemented here.

## Notes / follow-ups
- `hostnames` uses `shutter.gov.blockful.io` (mirrors `ens.gov.blockful.io`) — adjust if the real domain differs.
- `offchainProposals` intentionally **not** enabled: the gateway has no offchain endpoint for Shutter yet (would render a broken Offchain tab).
[...]

PR #2043, opened 2026-07-09
feat(dashboard): Shutter whitelabel rebrand — blue #0044A4 + new shield icon

brunod-e:
## What

Updates the Shutter whitelabel branding to their dark blue and new icon:

- **Brand color**: `svgColor` changed from `#051016` (near-black navy) to `#0044A4` in `shared/dao-config/shu.ts`. Propagates to all whitelabel theme tokens (`--base-brand`, links, buttons, surfaces) and the OG image background tint.
- **New icon**: `ShutterIcon` (app UI) replaced with the new striped-shield mark on a `#0044A4` chip; `ShutterOgIcon` (OG images) replaced with the monochrome variant that inherits the brand color like the other DAO OG icons (it previously hardcoded `fill="#EFEFEF"`, which made the OG art nearly invisible).
- **Theme contrast guard**: the whitelabel text-token guard in `shared/utils/theme.ts` (added for the old near-black navy) also washed out the new blue (`#0044A4` vs body text ≈ 2:1 < 3:1), rendering links/active states as a lightened mix. It now only lightens **near-neutral** dark brands (chroma < 48); saturated dark brands keep the pure color — they stay distinguishable from body text by hue and `#0044A4` has 8.9:1 contrast against the background.

## Verification

- OG image rendered locally: shield mark + oversized art in `#0044A4` over the light blue tint, matching the ENS OG style (the SVG `<mask>` clips correctly under Satori).
- Whitelabel UI driven via Playwright: sidebar active item, "All Proposals" chip and "New Proposal" button compute `rgb(0, 68, 164)` exactly; main dashboard (`/shu`, dark bg) renders the new icon in hero/sidebar/stage badge.
- `pnpm dashboard typecheck` and `pnpm dashboard lint` pass; create-proposal + shutter test suites pass (66/66).

## Note on the Slack unfurl (investigation)

The "Overview – Vercel" unfurl for `shutter.gov.blockful.io/proposals` is a stale Slack cache from before the domain was live — production currently serves correct OG meta and a valid image (verified with the Slackbot user agent). Sharing the URL with any query string (e.g. `?v=1`) fetches the fresh unfurl.


isadorable-png (commented):
## 🎨 UI Review

> Automated review · ℹ️ Reviewed without a Figma reference — searched the PR body/comments, the two preconfigured ClickUp lists, and a workspace-wide ClickUp search. The closest match, `DEV-960 "Shutter whitelabel"` (assigned to the PR author, status "code review"), links back to the original whitelabel PR (#1979) but its own 🕸️ Figma custom field is empty. No image attachments either. Grounded in diff/code analysis + UX-expert evaluation — the main dashboard Vercel preview shows "Ignored" in the bot comment (root-directory scoping) and no follow-up preview URL has posted yet, so no live visual confirmation was possible; only the Storybook deploy is reachable, which doesn't carry a Shutter icon story.

---

### Brand color migration — `shu.ts`, `theme.ts`

- Verified the new contrast/chroma guard by hand: `contrastRatio(#0044A4, #18181B)` (light/whitelabel theme body text) ≈ **2.0:1** — confirms the PR's claim that the raw contrast check alone would've triggered the lightened mix. `chroma(#0044A4) = 164`, well above the new `MIN_CHROMA = 48` threshold, so the guard correctly keeps the pure brand color for `--color-link` / `--color-link-hover` / `--color-highlight`. Math checks out — no defect found here.
- One thing worth a visual pass once the preview is reachable: the guard's chroma check is a proxy for "distinguishable from body text by hue," but it doesn't verify contrast of the link color *against the light-theme background* independent of body text — worth eyeballing `#0044A4` links/highlights on the whitelabel light theme once deployed, just to confirm they don't read as too "quiet" next to the near-black `#18181B` body copy despite the hue difference. [UX-heuristic]

---

### `ShutterIcon.tsx` / `ShutterOgIcon` — striped-shield mark

- `daoConfig.icon` renders at `size-8` (32px) in `WhitelabelSidebar.tsx` and `size-9` (36px) in `WhitelabelHeaderMobile.tsx` — confirmed via code search. The new mark's 5 horizontal stripes taper from ~6 units to ~2.4 units tall in the 68-unit viewBox — the thinnest bar is ≈1.1px at 32px render size. Vector rendering will still draw it, but sub-pixel strokes this thin commonly anti-alias into a faint/blurry line rather than a crisp stripe at small sizes. Worth a quick zoom-in check on the live sidebar/mobile header once the dashboard preview deploys — if the bottom stripe reads as "smudged" rather than a clean taper, that's the fix target (nothing to change in code without seeing it render). [Code-only]
- Good catch fixing `ShutterOgIcon`'s previously-hardcoded `fill="#EFEFEF"` to accept a `color` prop with `DEFAULT_FILL` fallback — now consistent with the other DAO OG icons (e.g. `FluidOgIcon`) in the same file.
- Background chip now fills the full `68×68` canvas with the pure brand color (`#0044A4`) instead of the old lightened `#12181C` chip — matches the PR's intent (saturated color doesn't need lightening) and is consistent with the theme-guard change above.

---

- The `showBackground` rect + shield mask approach is unchanged in structure from the previous icon (just re-authored path data at 68×68 instead of 40×40 with an equivalent centered margin) — scales correctly through the existing `width="100%" height="100%"` + `viewBox` pattern, no aspect-ratio or clipping regression expected against the `rounded-full` wrapper class used at both call sites.

---

pikonha (approved):

PR #2044, opened 2026-07-10
feat(user-api): User API — platform accounts & self-service API keys (backend)

brunod-e:
## What

The **entire backend** for platform accounts, consolidated into one PR (was #2044 + #2046 + #2047):

### New service: `apps/user-api` (:4003)
- **better-auth** sessions: SIWE (DB-backed single-use nonces, EIP-1271/ERC-6492 via `publicClient.verifyMessage`), magic link (Resend, env-gated) and Google OAuth (env-gated). No account linking in v1 — wallet and email/Google logins are separate users by design.
- **Per-host auth instances** derived from `AUTH_SIWE_DOMAINS` (one list drives SIWE domain, baseURL, cookies/CSRF and trusted origins) — no `BETTER_AUTH_URL`; whitelabel domains just get added to the list. The dashboard proxy forwards the browser host via `x-anticapture-host` (Railway's edge overwrites `x-forwarded-*`).
- **`GET /auth/methods`**: public capability discovery so the frontend only offers sign-in methods this deployment actually serves.
- **User-scoped drafts**: identity from the session, `daoId` column filter, server-generated ids, public share GET with server-derived `isOwner`, claim-on-first-SIWE-login for migrated rows, 100 drafts/user cap.
- **Self-service API keys** (DEV-950): minted through Authful under a per-user `user:<id>` tenant, unbounded rate (`rateLimitPerMin: 0` — free & unlimited for now), plaintext shown once, quota of 10 keys/user, `lastUsedAt` enrichment best-effort.

### Breaking: `apps/api`
- Draft-proposal routes/service/repository/mapper and `general-schema` **removed** — drafts belong to the User API now. One-shot migration script (`apps/user-api/scripts/migrate-drafts.ts`, `--dry-run` supported) preserves ids and lowercases author addresses.

### `apps/authful`
- **Scoped provisioning key**: may only mint/revoke `user:*` tenants and list a single `user:*` tenant via `?tenant=`; admin key unrestricted. Timing-safe key compare; foreign tenants answer 404 (no oracle).

### Infra
- `infra/user-api` Dockerfile + railway.json (migrate-then-serve entrypoint, `/health` healthcheck); `scripts/dev.sh` lane so `pnpm dev` boots the user-api locally.

## Architecture note
The User API is reached **only via the dashboard's same-origin `/api/user` proxy** (cookie sessions) — it does not sit behind Gateful. Bearer/API-key traffic keeps flowing through Gateful; user-api → Authful runs over the Railway private network.

## Testing
- 30 integration/unit tests green (PGlite + drizzle pushSchema; offline EOA verifier).
- Deployed to Railway dev and exercised E2E through the local dashboard proxy: SIWE nonce → verify → session cookie → create key (201 + plaintext) → list (no plaintext) → revoke (204).

Frontend counterpart: the stacked dashboard PR (base = this branch).
Replaces #2046 and #2047, which are closed in favor of this consolidation.


isadorable-png (commented):
## 🎨 UI Review

> Automated review · scope check only

This PR adds a new backend-only service, `apps/user-api` (better-auth SIWE/magic-link/Google sessions, user-scoped drafts — controllers, repositories, services, database schema, integration tests). No `.tsx`/component/style files, no dashboard integration in this PR — the dashboard counterpart ships separately in #2045. No rendered UI surface. Out of scope for a UI review; no findings to report.

---

isadorable-png:
## 🎨 UI Review

> Automated review · scope check only

This PR adds a new backend-only service, `apps/user-api` (SIWE, magic link & Google auth backend, session cookies, and a user-scoped drafts API). All 35 changed files are under `apps/user-api/**`, plus `.changeset/user-api-service.md` and `pnpm-lock.yaml` — no `.tsx`/component/style files, and no dashboard integration in this PR (the frontend counterpart is a stacked PR from `feat/user-accounts-dashboard`).

No rendered UI surface. Out of scope for a UI review; no findings to report.

---

apps/user-api/README.md
```diff
@@ -0,0 +1,53 @@
...
+
+## Env
+
+See `.env.example`. Required: `DATABASE_URL`, `BETTER_AUTH_SECRET`,
+`BETTER_AUTH_URL`, `AUTH_SIWE_DOMAINS`, `TRUSTED_ORIGINS`, `RPC_URL`.
+
+> `BETTER_AUTH_URL` is the dashboard **origin** (e.g. `https://dev.anticapture.com`),
+> not the `/api/user` proxy path. better-auth mounts at `baseURL` + `basePath`
+> (`/api/auth`), and the proxy strips `/api/user` before forwarding — so a path
+> in `BETTER_AUTH_URL` makes better-auth 404 every route.
```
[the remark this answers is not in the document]
brunod-e: Addressed in 0bb8e90c7: better-auth's browser-facing URLs (magic-link verify, OAuth callback) are served by a dashboard rewrite (`/api/auth/:path*` → the `/api/user` proxy), and magic-link emails point at the dashboard's `/auth/magic-link` interstitial. Documented in this README's "Browser-facing URLs" section; the rewrite + interstitial ship in #2051, which deploys together with this PR.

apps/user-api/package.json
```diff
@@ -0,0 +1,52 @@
...
+    "dev": "tsx watch --import ./src/instrumentation.ts src/index.ts",
+    "start": "node --import ./dist/instrumentation.js dist/index.js",
+    "build": "tsup",
+    "typecheck": "tsc --noEmit",
+    "test": "vitest run",
+    "test:unit": "vitest run",
+    "lint": "eslint src",
+    "lint:fix": "eslint src --fix",
+    "clean": "rm -rf node_modules *.tsbuildinfo dist",
+    "auth:generate": "better-auth generate --config ./src/auth.ts --output ./src/database/auth-schema.ts --yes",
```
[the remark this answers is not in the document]
brunod-e: Fixed in 0bb8e90c7 — `auth:generate` now points at `src/auth-instance.ts`, where the introspectable instance lives.

apps/user-api/src/services/api-keys/index.ts
```diff
@@ -0,0 +1,91 @@
...
+  constructor(
+    private readonly repo: ApiKeysRepository,
+    private readonly authful: AuthfulClient,
+    private readonly maxKeysPerUser = DEFAULT_MAX_KEYS_PER_USER,
+  ) {}
+
+  async create(userId: string, label: string): Promise<CreatedApiKey> {
+    const current = await this.repo.countActiveByUser(userId);
+    if (current >= this.maxKeysPerUser) {
+      throw new ApiKeyQuotaExceededError(this.maxKeysPerUser);
```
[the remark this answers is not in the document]
brunod-e: Fixed in 0bb8e90c7 — the insert re-checks the quota inside a per-user advisory-lock transaction (`createWithinQuota`); a create that loses the race to the last slot revokes its freshly minted Authful token and returns the quota error.

apps/user-api/src/controllers/drafts/index.ts
```diff
@@ -0,0 +1,225 @@
...
+      path: "/drafts",
+      summary: "Create a draft",
+      description:
+        "The draft id is generated server-side (it doubles as the share " +
+        "capability) and the author is the session user.",
+      middleware: [auth] as const,
+      request: {
+        body: {
+          content: { "application/json": { schema: CreateDraftBodySchema } },
+        },
```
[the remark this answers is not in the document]
brunod-e: Fixed in 0bb8e90c7 — the create and update bodies are `required: true`, so a body-less request fails validation with a 400.

apps/user-api/src/email/magic-link.ts
```diff
@@ -0,0 +1,53 @@
...
+ * The verify endpoint consumes its single-use token on the first GET, and
+ * mail pipelines routinely prefetch links (AV scanners, client previews,
+ * click-tracking) — a direct link would arrive already burned when the user
+ * clicks. The email therefore links the dashboard's interstitial page, which
+ * triggers the verify from the browser (JS/user gesture) that prefetchers
+ * don't execute.
+ */
+export const magicLinkLandingUrl = (verifyUrl: string): string => {
+  const verify = new URL(verifyUrl);
+  const landing = new URL("/auth/magic-link", verify.origin);
```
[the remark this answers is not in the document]
brunod-e: The interstitial ships in the frontend half of this stack (#2051, `apps/dashboard/app/auth/magic-link/page.tsx`) — the two PRs deploy together, and magic link stays env-gated off (no RESEND_API_KEY) until both are live. Called out in the README's "Browser-facing URLs" section.

apps/api/cmd/index.ts
```diff
@@ -363,15 +347,6 @@ votes(
...
-draftProposals(
-  app,
-  wrapWithTracing(
-    new DraftProposalsService(
-      wrapWithTracing(new DraftProposalsRepository(pgGeneralClient)),
-    ),
-  ),
-  env.DAO_ID.toLowerCase(),
-);
 docs(app);
```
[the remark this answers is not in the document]
brunod-e: Intentional: this is the breaking half of a two-PR stack — #2051 moves the dashboard's drafts to the User API and deploys together with this PR. Keeping the DAO-API draft routes alive would leave two write paths for the same data during the migration window.

apps/user-api/src/controllers/api-keys/index.ts
```diff
@@ -0,0 +1,148 @@
...
+      path: "/me/api-keys",
+      summary: "Create an API key",
+      description:
+        "Mints a key scoped to the session user via Authful. The plaintext is " +
+        "returned exactly once and never stored here.",
+      middleware: [auth] as const,
+      request: {
+        body: {
+          content: { "application/json": { schema: CreateApiKeyBodySchema } },
+        },
```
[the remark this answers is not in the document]
brunod-e: Fixed in 845d6986e — the body is `required: true` now, matching the drafts routes.

apps/api/cmd/index.ts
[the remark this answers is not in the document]
brunod-e: Follow-up: CI made your point concretely — the Vercel preview of this branch builds the (pre-accounts) dashboard against a client generated from this branch's gateful spec, and the missing draft exports broke the build. The removal moved to #2051 as you suggested (a3e39ec6c here re-adds the routes; the frontend PR retires them together with the dashboard cutover, so no caller exists in any deploy order).

apps/user-api/scripts/migrate-drafts.ts
pikonha, line 1: I wouldn't commit this given that it will be used only once

scripts/dev.sh
```diff
@@ -247,6 +249,22 @@ else
...
+  log "Starting optional User API..."
+  run_with_prefix "$C_USER_API" "👤 user-api" "" "" railway run -e dev -s user-api pnpm user-api dev &
+  if wait_for_optional_port "$PORT_USER_API" "User API"; then
+    USER_API_AVAILABLE=true
+    export USER_API_URL="http://localhost:${PORT_USER_API}"
+  fi
+else
+  log "Skipping optional User API (Railway CLI/service unavailable)"
+fi
+
```
pikonha, line 267: can't you just run it locally?

apps/user-api/package.json
```diff
@@ -0,0 +1,53 @@
...
+  "type": "module",
+  "main": "dist/index.js",
+  "description": "User API — user identity & session service (SIWE, Google, magic link) and off-chain user data (drafts)",
+  "scripts": {
+    "dev": "tsx watch --import ./src/instrumentation.ts src/index.ts",
+    "start": "node --import ./dist/instrumentation.js dist/index.js",
+    "build": "tsup",
+    "typecheck": "tsc --noEmit",
+    "test": "vitest run",
+    "test:unit": "vitest run",
```
pikonha:
```suggestion
    "test": "vitest run",
```

apps/user-api/package.json
```diff
@@ -0,0 +1,53 @@
...
+    "lint": "eslint src",
+    "lint:fix": "eslint src --fix",
+    "clean": "rm -rf node_modules *.tsbuildinfo dist",
+    "auth:generate": "better-auth generate --config ./src/auth-instance.ts --output ./src/database/auth-schema.ts --yes",
```
pikonha, line 17: what is this used for?

apps/user-api/drizzle.config.ts
```diff
@@ -0,0 +1,17 @@
...
+// unlike the DAO API's namespaced `general`/`anticapture` schemas — no
+// schemaFilter is needed and the better-auth CLI's unqualified table output
+// works directly.
+export default defineConfig({
+  schema: "./src/database/schema.ts",
+  out: "./drizzle",
+  dialect: "postgresql",
+  dbCredentials: {
+    url: process.env.DATABASE_URL!,
+  },
```
pikonha, line 16: this avoind hanging connections with no timeout

```suggestion
  connection: {
    connectionString: process.env.DATABASE_URL!,
    connectionTimeoutMillis: 10_000,
    statement_timeout: 30_000,
  },
```

apps/user-api/.env.example
```diff
@@ -0,0 +1,35 @@
+# User API — user identity & session service
+
+PORT=4003
+
+# Dedicated Postgres for the User API (its own DB, not the DAO/general schema).
+DATABASE_URL=postgresql://postgres:admin@localhost:5432/user_api
+
+# Session signing key — must be >= 32 high-entropy chars.
```
pikonha:
```suggestion
# Session signing key — openssl rand -base64 32
```

apps/user-api/scripts/migrate-drafts.test.ts
pikonha, line 1: remove it

apps/user-api/src/controllers/api-keys/index.ts
```diff
@@ -0,0 +1,140 @@
...
+  type ApiKeysService,
+} from "@/services/api-keys";
+
+const toResponse = (row: ApiKeyRow, lastUsedAt: string | null = null) => ({
+  id: row.id,
+  label: row.label,
+  createdAt: row.createdAt.toISOString(),
+  revokedAt: row.revokedAt?.toISOString() ?? null,
+  lastUsedAt,
+});
```
pikonha, line 24: this could be parsing with the zod schema (`ApiKeySchema.parse()`)


```suggestion
```

apps/user-api/src/controllers/api-keys/index.ts
```diff
@@ -0,0 +1,140 @@
...
+        ...unauthorizedResponses,
+      },
+    }),
+    async (c) => {
+      const { id: userId } = c.get("sessionUser");
+      const keys = await service.list(userId);
+      return c.json(
+        ApiKeyListResponseSchema.parse({
+          items: keys.map((k) => toResponse(k, k.lastUsedAt)),
+        }),
```
pikonha, line 54:
```suggestion
        ApiKeyListResponseSchema.parse({
          items: keys,
        }),
```

apps/user-api/src/controllers/api-keys/index.ts
```diff
@@ -0,0 +1,140 @@
...
+          CreatedApiKeyResponseSchema.parse({
+            ...toResponse(key),
+            token: plaintext,
+          }),
+          201,
+        );
+      } catch (err) {
+        if (err instanceof ApiKeyQuotaExceededError) {
+          return c.json({ error: "api_key_limit_reached" }, 403);
+        }
```
pikonha, line 107: this is cool

apps/user-api/src/controllers/api-keys/index.ts
```diff
@@ -0,0 +1,140 @@
...
+      },
+    }),
+    async (c) => {
+      const { label } = c.req.valid("json");
+      const { id: userId } = c.get("sessionUser");
+      try {
+        const { key, plaintext } = await service.create(userId, label);
+        return c.json(
+          CreatedApiKeyResponseSchema.parse({
+            ...toResponse(key),
```
pikonha, line 99: again, why do you need this instead of relying on the zod schema

apps/user-api/src/controllers/drafts/index.ts
```diff
@@ -0,0 +1,218 @@
...
+  title: row.title,
+  discussionUrl: row.discussionUrl,
+  body: row.body,
+  actions: row.actions as Record<string, unknown>[],
+  createdAt: row.createdAt,
+  updatedAt: row.updatedAt,
+  // Never derived client-side: a Google author has no wallet to compare, and
+  // the caller's identity comes exclusively from the session.
+  isOwner: row.userId !== null && row.userId === sessionUserId,
+});
```
pikonha, line 33: rely on the zod schema instead of manual mapper

apps/user-api/src/controllers/drafts/index.ts
```diff
@@ -0,0 +1,218 @@
...
+  service: DraftsService,
+  resolver: AuthResolver,
+) {
+  const auth = sessionAuth(resolver);
+  const maybeAuth = optionalSession(resolver);
+
+  app.openapi(
+    createRoute({
+      method: "get",
+      operationId: "listDrafts",
```
pikonha:
```suggestion
      operationId: "listProposalDrafts",
```

apps/user-api/src/controllers/drafts/index.ts
```diff
@@ -0,0 +1,218 @@
...
+      path: "/drafts",
+      summary: "List the session user's drafts for a DAO",
+      middleware: [auth] as const,
+      request: { query: ListDraftsQuerySchema },
+      responses: {
+        200: {
+          description: "Drafts owned by the session user, newest first",
```
pikonha:
```suggestion
          description: "Proposal drafts owned by the session user, newest first",
```

pikonha (changes requested):
isn't this supposed to remove the drafts from the APIs?

.github/workflows/tests.yaml
pikonha, line 1: why can't you just use the one from the PR environment? no need for interpolation

revert all this file

apps/user-api/src/services/drafts/index.ts
```diff
@@ -0,0 +1,97 @@
...
+export const DEFAULT_MAX_DRAFTS_PER_USER = 100;
+
+export class DraftQuotaExceededError extends Error {
+  constructor(readonly limit: number) {
+    super(`draft limit of ${limit} reached`);
+    this.name = "DraftQuotaExceededError";
+  }
+}
+
+export class DraftsService {
```
pikonha: drafts is too general

```suggestion
export class ProposalDraftsService {
```

apps/user-api/src/services/drafts/index.ts
```diff
@@ -0,0 +1,97 @@
...
+   */
+  private readonly claimed = new Set<string>();
+
+  /**
+   * Adopts migrated rows (userId NULL) authored by one of this user's
+   * wallets — claim-on-first-touch, so ownership is settled on EVERY entry
+   * point (list, direct share view, update, delete), not just the list.
+   * Idempotent. Returns the user's wallets for callers that need them.
+   */
+  private async claimMigrated(userId: string): Promise<string[]> {
```
pikonha: not sure I understand what `claim` means on this context

apps/user-api/src/services/drafts/index.ts
```diff
@@ -0,0 +1,97 @@
...
+      await this.repo.claimByAddresses(userId, wallets);
+      // Guard against unbounded growth across a very long process life.
+      if (this.claimed.size >= 10_000) this.claimed.clear();
+      this.claimed.add(userId);
+    }
+    return wallets;
+  }
+
+  async listForUser(userId: string, daoId: string): Promise<DraftRow[]> {
+    await this.claimMigrated(userId);
```
pikonha, line 41: what's  this for?

apps/user-api/src/email/magic-link.ts
```diff
@@ -0,0 +1,53 @@
...
+        <div style="font-family: system-ui, sans-serif; max-width: 480px; margin: 0 auto;">
+          <h2 style="font-size: 18px;">Sign in to Anticapture</h2>
+          <p style="color: #555;">Click the button below to sign in. This link expires shortly and can be used once.</p>
+          <p style="margin: 24px 0;">
+            <a href="${landingUrl}" style="background: #E66AE9; color: #fff; padding: 12px 20px; border-radius: 8px; text-decoration: none; display: inline-block;">Sign in</a>
+          </p>
+          <p style="color: #888; font-size: 13px;">If you didn't request this, you can safely ignore this email.</p>
+        </div>
+      `,
+    });
```
pikonha, line 48: isn't it gonna be reviewed by @isadorable-png ?

apps/user-api/src/middlewares/logger.ts
```diff
@@ -0,0 +1,34 @@
...
+    }
+
+    logger.info(
+      {
+        method: c.req.method,
+        path: c.req.path,
+        status: c.res.status,
+        durationMs: Math.round(performance.now() - start),
+      },
+      "request completed",
```
pikonha:
```suggestion
     `${c.req.method} ${c.req.path} ${c.res.status}`,
```

apps/user-api/src/middlewares/logger.ts
```diff
@@ -0,0 +1,34 @@
...
+      await next();
+    } catch (err) {
+      logger.error(
+        {
+          err,
+          method: c.req.method,
+          path: c.req.path,
+          durationMs: Math.round(performance.now() - start),
+        },
+        "request failed",
```
pikonha:
```suggestion
        `${c.req.method} ${c.req.path} ${c.res.status}`,
```

apps/authful/src/index.ts
```diff
@@ -42,6 +42,7 @@ const app = createApp({
   db,
   adminApiKey: env.ADMIN_API_KEY,
   internalApiKey: env.INTERNAL_API_KEY,
+  provisioningApiKey: env.PROVISIONING_API_KEY,
```
pikonha, line 52: what's this used for? couldn't it use the same keys it already had? i see 3 different keys now

apps/user-api/src/database/index.ts
```diff
@@ -0,0 +1,9 @@
+import { drizzle } from "drizzle-orm/node-postgres";
+
+import { env } from "@/env";
+import * as schema from "./schema";
+
+export const db = drizzle(env.DATABASE_URL, { schema });
```
pikonha:
```suggestion
export const db = drizzle({
  connection: {
    connectionString: env.DATABASE_URL,
    connectionTimeoutMillis: 10_000,
    statement_timeout: 30_000,
  },
   schema
});
```

apps/user-api/drizzle/0000_lazy_lake.sql
pikonha, line 1: have you created this SQL manually or drizzle generated it?

.claude/skills/local-dev-stack/SKILL.md
```diff
@@ -46,6 +46,7 @@ the first.
 | Dashboard          | 3000  | http://localhost:3000  | always                              |
 | Address Enrichment | 3001  | http://localhost:3001  | optional, skipped if no Railway svc |
 | Relayer (ENS)      | 3002  | http://localhost:3002  | ENS only                            |
+| User API           | 4003  | http://localhost:4003  | optional, skipped if no Railway svc |
```
pikonha:
```suggestion
| User API           | 4003  | http://localhost:4003  | optional |
```

apps/user-api/src/clients/authful/index.ts
```diff
@@ -0,0 +1,76 @@
...
+
+  private headers() {
+    return {
+      Authorization: `Bearer ${this.provisioningApiKey}`,
+      "Content-Type": "application/json",
+    };
+  }
+
+  async mint(tenant: string, name: string): Promise<MintedToken> {
+    const res = await fetch(`${this.baseUrl}/tokens`, {
```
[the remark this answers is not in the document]
brunod-e: Fixed in 5b1c3c33e — the client constructor now trims trailing slashes off AUTHFUL_URL (same trim gateful applies to its TOKEN_SERVICE_URL), so `//tokens`-style paths can't happen.

apps/user-api/src/services/api-keys/index.ts
```diff
@@ -0,0 +1,100 @@
...
+    // authoritative — the quota is enforced atomically at insert time.
+    const current = await this.repo.countActiveByUser(userId);
+    if (current >= this.maxKeysPerUser) {
+      throw new ApiKeyQuotaExceededError(this.maxKeysPerUser);
+    }
+
+    // Mint in Authful under this user's own tenant, then record ownership.
+    const minted = await this.authful.mint(
+      `${USER_TENANT_PREFIX}${userId}`,
+      label,
```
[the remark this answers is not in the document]
brunod-e: Fixed in 5b1c3c33e, on both sides: user-api mints with a bounded constant name ("self-service") — the human label lives only in user-api's own table — and gateful's usage middleware buckets `user:*` tenants as `user:*` in `tenantRequestTotal`. Neither Prometheus label is user-controlled or unbounded anymore (gateful changeset included).

apps/user-api/src/env.ts
```diff
@@ -0,0 +1,79 @@
...
+    // Magic link: enabled when RESEND_API_KEY is set. RESEND_FROM_EMAIL
+    // defaults to Resend's sandbox sender for local/dev.
+    RESEND_API_KEY: z.string().optional(),
+    RESEND_FROM_EMAIL: z.string().default("onboarding@resend.dev"),
+
+    // User self-service API keys: enabled when both are set. The User API
+    // brokers keys into Authful over the private network using a provisioning
+    // key scoped to `user:*` tenants.
+    AUTHFUL_URL: z.string().url().optional(),
+    AUTHFUL_PROVISIONING_API_KEY: z.string().min(16).optional(),
```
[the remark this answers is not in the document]
brunod-e: Fixed in 5b1c3c33e — blank AUTHFUL_URL / AUTHFUL_PROVISIONING_API_KEY normalize to `undefined` before validation (same union pattern as authful's `optionalSecret`), so an empty env-template value keeps the feature disabled instead of failing a SIWE-only boot.

.github/workflows/tests.yaml
```diff
@@ -89,20 +93,27 @@ jobs:
...
+              --header "Content-Type: application/json" \
+              --data "${payload}"
+
+            echo "Configured $1=$2 for Vercel preview branch ${PR_BRANCH}."
+          }
+
+          # ANTICAPTURE_API_URL feeds the data proxy (/api/gateful);
+          # USER_API_URL feeds the auth/session proxy (/api/user).
+          upsert "ANTICAPTURE_API_URL" "${GATEFUL_URL}"
+          upsert "USER_API_URL" "${USER_API_URL}"
```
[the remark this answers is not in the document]
brunod-e: Fixed in 5b1c3c33e — `user_api_url` is now a job output and the CLI preview deploy passes `--env USER_API_URL` alongside ANTICAPTURE_API_URL (runtime-only: only the /api/user proxy reads it at request time, nothing needs it at build).

apps/user-api/src/services/api-keys/index.ts
```diff
@@ -0,0 +1,100 @@
...
+   * Revokes a key the user owns. Revokes in Authful first (the security-
+   * relevant step) so a failure there aborts before we mark it revoked
+   * locally — never the reverse, which would leave a live token the user
+   * believes is gone. Returns false if the key isn't found or isn't theirs.
+   */
+  async revoke(id: string, userId: string): Promise<boolean> {
+    const key = await this.repo.findOwned(id, userId);
+    if (!key || key.revokedAt) return false;
+
+    await this.authful.revoke(key.authfulTokenId);
```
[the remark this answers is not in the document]
brunod-e: Fixed in 5b1c3c33e — gateful now caches positive verdicts for `user:*` tenants for 30s (ops tenants keep the outage-tolerant 300s), so a dashboard revoke stops authenticating in seconds rather than minutes. Changeset included.

apps/user-api/src/services/drafts/index.ts
```diff
@@ -0,0 +1,97 @@
...
+   * point (list, direct share view, update, delete), not just the list.
+   * Idempotent. Returns the user's wallets for callers that need them.
+   */
+  private async claimMigrated(userId: string): Promise<string[]> {
+    const wallets = await this.repo.findWalletAddresses(userId);
+    if (!this.claimed.has(userId)) {
+      await this.repo.claimByAddresses(userId, wallets);
+      // Guard against unbounded growth across a very long process life.
+      if (this.claimed.size >= 10_000) this.claimed.clear();
+      this.claimed.add(userId);
```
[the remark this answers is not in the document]
brunod-e: Fixed in 5b1c3c33e — the process-local memo is gone and claim-on-touch is idempotent per request again, precisely because the one-shot migration can insert claimable rows after a user was first seen. (The memo was a recent perf tweak; correctness wins.)

apps/user-api/scripts/migrate-drafts.ts
brunod-e: It can't run before the prod cutover happens, so deleting it now would orphan the migration plan: this is the documented, tested path that moves `general.proposal_drafts` from each DAO DB into the user-api when the accounts release ships. Plan: run it once per DAO at cutover, then delete script + test in the cleanup PR. If you'd rather keep the tree clean now and run it from a branch at cutover time, happy to remove — your call.

scripts/dev.sh
brunod-e: It does run locally — the process is a plain `pnpm user-api dev` on your machine; only the env comes from `railway run` (its dedicated Postgres lives in the Railway dev env, same pattern as address-enrichment above). Fully-local works too: point DATABASE_URL at a local Postgres in apps/user-api/.env and start it directly — the Railway path is just the zero-setup default.

apps/user-api/package.json
brunod-e: Applied in 5b1c3c33e — the duplicate `test:unit` script is gone (`test` was already `vitest run`; turbo skips packages without the script).

apps/user-api/package.json
brunod-e: `auth:generate` runs the better-auth CLI against src/auth-instance.ts to regenerate src/database/auth-schema.ts — the drizzle schema for better-auth's own tables (user/session/account/verification/walletAddress). It's only needed when the auth config changes (e.g. adding a plugin); the generated file is committed, and db:generate then derives SQL migrations from it.

apps/user-api/drizzle.config.ts
brunod-e: Applied where the driver accepts it (5b1c3c33e): the runtime pool in src/database/index.ts now sets connectionTimeoutMillis/statement_timeout (your other suggestion). drizzle-kit's `dbCredentials` only takes url/ssl — no pool options — so the config file stays as-is; the kit CLI opens short-lived interactive connections anyway.

apps/user-api/.env.example
brunod-e: Applied in 5b1c3c33e.

apps/user-api/scripts/migrate-drafts.test.ts
brunod-e: Kept together with the script for now (see the thread on migrate-drafts.ts) — the test pins the untrusted-DB-content validation (malformed ids/addresses must not cross into the new table) until the cutover runs; both get deleted in the same cleanup PR afterwards.

apps/user-api/src/controllers/api-keys/index.ts
brunod-e: The zod schema is already the validator — every handler returns `Schema.parse(...)`. `toResponse` exists for what the schema deliberately doesn't encode: drizzle rows carry `Date` objects while the wire contract is ISO strings (`createdAt: z.string()`), plus the lastUsedAt default. Folding that into the schema would need input-unions/transforms that leak DB types into the generated OpenAPI. If you'd rather see `z.codec`/transform schemas here, can do — but the spec output gets uglier.

apps/user-api/src/controllers/api-keys/index.ts
brunod-e: `ApiKeyListResponseSchema.parse({ items: keys })` throws at runtime: `keys` rows carry `Date` objects for createdAt/revokedAt while the schema (wire contract) declares ISO strings — the mapper is the Date→ISO seam, parse stays the validator. Same rationale as the thread above.

apps/user-api/src/controllers/api-keys/index.ts
brunod-e: 🙏

apps/user-api/src/controllers/api-keys/index.ts
brunod-e: Same seam as above: the schema validates the wire shape; the mapper converts driver types (Date→ISO). Without it, parse() rejects the rows.

apps/user-api/src/controllers/drafts/index.ts
brunod-e: Same as the api-keys threads — `DraftListResponseSchema.parse` is already applied to every response; the mapper's job is the type conversion (and deriving `isOwner` from the session, which no schema can know).

apps/user-api/src/controllers/drafts/index.ts
brunod-e: Applied in 5b1c3c33e.

apps/user-api/src/controllers/drafts/index.ts
brunod-e: Applied in 5b1c3c33e.

.github/workflows/tests.yaml
brunod-e: Partially applied in 5b1c3c33e, but a plain revert would kill preview auth. What the interpolation is for: each trusted PR gets its own Railway preview env, whose service URLs embed the PR number (`user-api-anticapture-pr-N.up.railway.app`) — there's no static value 'from the PR environment' that the Vercel preview could read instead (Vercel can't see Railway's env). The trust gate exists so fork PRs never get preview credentials and fall back to shared dev. What changed in 5b1c3c33e: the deploy step now receives the computed URL directly via `--env` (codex's point — branch-scoped envs don't apply to CLI deploys); the upsert stays because Vercel's own git-triggered preview deploys DO read branch-scoped envs. If you want, we can pair on simplifying the file, but each piece currently has a consumer.

apps/user-api/src/services/drafts/index.ts
brunod-e: Applied in 5b1c3c33e — `ProposalDraftsService` (rename propagated through app wiring and tests).

apps/user-api/src/services/drafts/index.ts
brunod-e: Fair — the docstring now spells it out (5b1c3c33e): legacy drafts were migrated from the per-DAO DBs with `user_id NULL` and only the author's wallet on them (accounts didn't exist). "Claiming" = attaching those still-unowned rows to the signed-in user whose wallet matches, on first touch.

apps/user-api/src/services/drafts/index.ts
brunod-e: It records the author's wallet on drafts created via the API: shared drafts display the author address publicly, and it matches the claim key shape used by migrated rows. Email/Google authors simply have none (null). Comment clarified in 5b1c3c33e.

apps/user-api/src/email/magic-link.ts
brunod-e: Good call — flagging it for @isadorable-png. The current template is deliberately minimal/utilitarian; a design pass can change it freely without touching any contract (it's one function that receives the URL).

apps/user-api/src/middlewares/logger.ts
brunod-e: Applied in 5b1c3c33e (kept the structured fields for querying; the message line now reads `GET /path 200`).

apps/user-api/src/middlewares/logger.ts
brunod-e: Applied in 5b1c3c33e.

apps/authful/src/index.ts
brunod-e: Least privilege — the three keys have different blast radii: ADMIN_API_KEY is full token admin over every tenant (ops only); INTERNAL_API_KEY is gateful's validate-only credential; PROVISIONING_API_KEY (new) lets the user-api mint/revoke/list tokens but ONLY under `user:*` tenants — the handlers enforce the prefix. Reusing ADMIN_API_KEY would mean a compromised user-api (the most internet-exposed of the three consumers) could mint/revoke ops tenants' tokens. One key per caller also means one revocation per caller.

apps/user-api/src/database/index.ts
brunod-e: Applied in 5b1c3c33e.

apps/user-api/drizzle/0000_lazy_lake.sql
brunod-e: Generated — `pnpm auth:generate` (better-auth CLI) produced the drizzle schema for the auth tables, then `pnpm db:generate` (drizzle-kit) emitted this SQL from it. Not hand-written.

.claude/skills/local-dev-stack/SKILL.md
brunod-e: Applied in 5b1c3c33e.

apps/user-api/src/mappers/drafts/index.ts
```diff
@@ -0,0 +1,62 @@
...
+  })
+  .openapi("DraftParams");
+
+export const CreateDraftBodySchema = z
+  .object({
+    daoId: z.string().min(1),
+    title: z.string().default(""),
+    discussionUrl: z.string().default(""),
+    body: z.string().default(""),
+    actions: z.array(ActionSchema).default([]),
```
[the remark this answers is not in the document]
brunod-e: Fixed in 244edbe17 — every persisted field is bounded (title 300, discussionUrl 2048, body 100k, 50 actions, daoId 64) and the app adds a global 1 MiB body limit as backstop, so the row quota can't be turned into per-row megabytes. Covered by an oversized-title 400 test.

apps/authful/src/middlewares/token-auth.ts
```diff
@@ -0,0 +1,57 @@
...
+  adminApiKey: string;
+  provisioningApiKey?: string;
+}) =>
+  createMiddleware(async (c, next) => {
+    const token = bearer(c.req.header("Authorization"));
+    if (!token) return c.json({ error: "unauthorized" }, 401);
+
+    if (safeEqual(token, opts.adminApiKey)) {
+      c.set("authScope", "admin");
+      return next();
```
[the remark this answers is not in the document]
brunod-e: Fixed in 244edbe17 — authful's env validation now refuses to boot when PROVISIONING_API_KEY equals ADMIN_API_KEY or INTERNAL_API_KEY, so the admin-first match in scopedTokenAuth can never silently void the user:* restriction.

apps/user-api/src/controllers/api-keys/api-keys.integration.test.ts
```diff
@@ -0,0 +1,259 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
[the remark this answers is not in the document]
brunod-e: Fixed in 244edbe17 — the file-wide suppression is gone. Responses are validated with the route schemas (CreatedApiKeyResponseSchema / ApiKeyListResponseSchema), the one leak assertion keeps a typed raw shape (zod parse would strip an accidental `token` field and defeat it), and the only remaining cast is the known drizzle-kit pushSchema typing gap with an inline-scoped disable — same pattern the drafts suite now uses.

apps/authful/src/repositories/tokens/index.ts
```diff
@@ -12,6 +12,14 @@ export class TokensRepository {
     return this.db.select().from(tokens).orderBy(desc(tokens.createdAt));
   }
 
+  async listByTenant(tenant: string): Promise<DBToken[]> {
+    return this.db
+      .select()
+      .from(tokens)
+      .where(eq(tokens.tenant, tenant))
```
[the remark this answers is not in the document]
brunod-e: Fixed in 244edbe17 — the provisioning scope's tenant listing now returns active tokens only (new listActiveByTenant), so the enrichment payload stays bounded by the key quota under churn; admin keeps the full-history view.

apps/user-api/src/auth.ts
```diff
@@ -0,0 +1,215 @@
...
+    if (existing) return existing;
+
+    // Preview envs only: every Vercel preview deployment gets a unique URL,
+    // so instances for them are built on demand. Anything that isn't a
+    // *.vercel.app host still fails closed, exactly like dev/production.
+    if (!config.previewDynamicHosts || !isPreviewHost(host)) return undefined;
+
+    const dynamic = createAuth(config, host);
+    if (instances.size < MAX_DYNAMIC_INSTANCES) instances.set(host, dynamic);
+    return dynamic;
```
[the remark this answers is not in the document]
brunod-e: Fixed in 244edbe17 — the resolver checks the cap before construction and fails closed (untrusted_host) at MAX_DYNAMIC_INSTANCES instead of rebuilding an uncached instance per request.

pikonha (approved):

PR #2045, opened 2026-07-10
feat(dashboard): platform accounts — sign-in modal (SIWE/email/Google) + drafts on the User API

brunod-e:
## Summary

Dashboard side of the platform-accounts architecture (stacked on #2044 — the `apps/user-api` service). Adds the sign-in modal from the [Figma design](https://www.figma.com/design/mUgy2KpQ3gJ07yZaUaXu8l/%F0%9F%9B%B0%EF%B8%8F-Product-Design?node-id=3315-19892) and moves draft proposals onto the session-scoped User API.

## Changes

- **`/api/user` proxy**: same-origin, cookie-relaying reverse proxy to the User API (every `Set-Cookie` preserved; forwards `Origin` + `x-forwarded-host`, which better-auth needs behind a path-rewriting proxy). Unlike the gateful proxy, no shared bearer is attached — identity is the user's own session.
- **Sign-in modal** (design ③.1/③.2): wallet (full SIWE ceremony via `useSiweLogin` — opens RainbowKit connect when needed, resumes after connect), email (magic link with enter-email → “Check your inbox” + 30s resend cooldown), Google (social sign-in). Email/Google render only where enabled (`NEXT_PUBLIC_EMAIL_LOGIN` / `NEXT_PUBLIC_GOOGLE_LOGIN`); **whitelabel is wallet-only**. Mounted app-wide via `LoginProvider` (`useLogin().openLogin()`).
- **Design system**: `Modal` gains an optional header — headerless modals render a bare close button + screen-reader-only name (`ariaLabel`), so bespoke layouts like the login surface reuse the same dialog.
- **Drafts on the User API**: hand-written `/api/user/drafts` client (the User API is not in gateful's spec, so no generated SDK). `useDrafts` is session-gated (`useSession`), sends no address params, and exposes `needsAuth`; `ProposalCreationForm` uses the server-derived `isOwner` for Editor-vs-Preview (correct for email/Google authors, who have no wallet), and Save prompts sign-in when unauthenticated. Removes dead `useDraftRecipient`.

## Verification

- `pnpm typecheck` clean; eslint clean on changed files; 63 create-proposal jest tests green.
- Manual E2E pending local User API Postgres (follow-up: dev-stack lane for the service).

## Notes

- The old gateful draft endpoints remain live until the follow-up PR removes them from the DAO APIs (contract change + client major).

Refs DEV-978, DEV-979.


isadorable-png:
## 🎨 UI Review

⚠️ Figma link found but Figma MCP unreachable — visual validation skipped, configuration error reported. This review is code-only + UX-heuristic; no finding below is validated against the actual Figma design. Vercel preview (`anticapture-storybook-git-feat-user-accounts-dashboard-ful.vercel.app`) returned **403 Forbidden** (deployment protection) — could not load the live preview either.

### Sign-in modal (`LoginModal.tsx`) — options screen

1. **[Code-only] Headerless close button loses the established 44px touch target.** In `Modal.tsx`, the header-present desktop dialog gives its close button an explicit hit-area override:
   ```tsx
   // ModalHeader.tsx
   <IconButton variant="ghost" size="sm" icon={X} aria-label="Close" className="-m-2 size-11" />
   ```
   but the new headerless branch (used by `LoginModal`, since it passes no `title`) omits that override:
   ```tsx
   // Modal.tsx, headerless branch
   <IconButton variant="ghost" size="sm" icon={X} aria-label="Close" className="absolute right-3 top-3 z-10" />
   ```
   Without `-m-2 size-11` (or equivalent), the button falls back to `IconButton`'s default `sm` box (`p-1` + `size-3.5` ≈ 22×22px) — about half the tap target used one branch away in the same file. Add `size-11 -m-2` (adjusting the `absolute` offsets accordingly) to match the DS convention already established in `ModalHeader.tsx`.

2. **[Code-only] Legal-copy link uses the wrong link variant.** The "Terms of Use & Privacy Policy" link at the bottom of the sign-in modal renders via `DefaultLink`, whose base classes are `flex items-center gap-1 font-mono tracking-wider uppercase leading-none font-medium` (`DefaultLink.tsx`). That's the nav/meta-link style (e.g. "VIEW ON ETHERSCAN"-type links), so this will render as small, tracked, monospace, ALL-CAPS text:
   ```tsx
   <DefaultLink href="/terms-of-service" openInNewTab size="sm" className="text-dimmed hover:text-secondary">
     Terms of Use & Privacy Policy
   </DefaultLink>
   ```
   That's an unusual treatment for an inline legal disclaimer sentence sitting under a centered login card. Recommend a plain inline text link (no `font-mono`/`uppercase`/`tracking-wider`) rather than reusing `DefaultLink` here.

### Error states (SIWE + email)

3. **[Code-only] Raw provider/server errors are shown to the user verbatim, with no friendly-copy mapping**, which is inconsistent with the otherwise polished copy elsewhere ("Check your inbox", the resend countdown):
   ```ts
   // useSiweLogin.ts
   setError(err instanceof Error ? err.message : "Sign-in failed");
   // useEmailLogin.ts
   setError(sendError.message ?? "Could not send the login link");
   ```
   A wallet rejection (e.g. viem/wagmi's `UserRejectedRequestError: User rejected the request.`) or a raw better-auth error string will surface directly in the modal. Suggest mapping known error codes to short user-facing copy (e.g. "Signature request was cancelled") and falling back to the generic string only for unknowns.

### Accessibility (minor)

4. **[UX-heuristic] Resend cooldown countdown isn't in an `aria-live` region.** `ResendButton` updates its label every second ("Resend in 30s" → … → "Resend link") purely via the disabled button's text; screen-reader users get no announcement when the button becomes actionable. Wrapping the label (or the button) in `aria-live="polite"` would close this gap. Low severity since it's a disabled control, but worth a one-line fix.

5. **[Code-only] "Back" is a hand-rolled `<button>`** with bespoke classes (`text-secondary hover:text-primary mx-auto flex items-center gap-1 text-sm transition-colors`) rather than a DS `Button`/link primitive. Minor DS-adherence nit — functionally fine, just not reusing existing primitives like the rest of the modal does.

### Mobile

6. **[UX-heuristic, low confidence] Mobile drawer keeps a full header band even in "headerless" mode.** On mobile, `Modal` renders via `DrawerHeader`, which still paints its `bg-surface-contrast` header row (with only an invisible `sr-only` title) above the centered logo/content. On desktop the headerless variant has no separate header band at all — just a bare floating close button over the body. That's a real compositional difference between breakpoints for the same modal; it may be intentional (bottom sheets in this codebase apparently always keep the header band for the close-button anchor), but I can't confirm intent without Figma this run — worth a quick visual check against the design.

---
*Automated UI review — Path B (Figma MCP unreachable this run). Findings are `[Code-only]`/`[UX-heuristic]`; none are `[Figma-confirmed]`.*

---

isadorable-png (commented):
## 🎨 UI Review

**Review basis (Path B — Figma link found, MCP unreachable):** The PR description links Figma node `3315-19892` (sign-in modal, states ③.1/③.2), but the Figma MCP connector needs an OAuth authorization this non-interactive session can't complete. Findings below are grounded in the diff, the design-system source (`Modal`, `Button`, `IconButton`, `DrawerHeader`), and UX heuristics — **not** a pixel-level Figma diff. Please cross-check against the Figma frames directly, especially #3 and #4 below.

**Preview:** Only the `anticapture-storybook` Vercel project deploys for this branch (the `apps/dashboard` project is "Ignored" per this repo's root-directory scoping quirk). The reachable storybook preview URL returned `403 Forbidden` (SAML-gated), so no live app preview could be checked either.

---

### Sign-in modal — `apps/dashboard/shared/components/auth/LoginModal.tsx`

**Must-fix**

1. **Save → sign-in flow silently drops the user's action.** [Code-only] `ProposalCreationForm.handleSaveDraft` calls `openLogin()` when `drafts.needsAuth` (`apps/dashboard/features/create-proposal/components/ProposalCreationForm.tsx` ~L311-317), but `LoginProvider` is mounted once globally with only `isWhitelabel` (`apps/dashboard/shared/providers/GlobalProviders.tsx` L43) and `useLogin()`'s `openLogin()` takes no resume callback. On successful sign-in, `LoginModal.tsx` L45-48 just does `onOpenChange(false); onAuthenticated?.()` — but nothing is ever passed as `onAuthenticated` at the global mount, so the modal closes and the save the user asked for never happens, with zero feedback. The user has to notice nothing happened and click Save again. Wire a resume callback through `openLogin(onSuccess)`, or at minimum toast "Signed in — click Save again" after auth.

2. **Headerless close button loses the 44px touch target.** [Code-only] `ModalHeader`'s close button explicitly widens its hit area to 44px (`className="-m-2 size-11"`, `apps/dashboard/shared/components/design-system/modal/modal-header/ModalHeader.tsx` L52). The new headerless branch added in `Modal.tsx` (~L146-152) renders the same `IconButton` at `size="sm"` with no size override (`className="absolute right-3 top-3 z-10"`), so it renders at its default ~22px box — this is the only way to close the sign-in modal on desktop, and it's half the sibling header's target and below the 44px minimum. Add the equivalent size override.

3. **Subtitle copy doesn't match which methods are actually shown.** [Code-only] The default view always renders "Use your wallet, or just your email." (`LoginModal.tsx` L145), independent of `showEmail`/`showGoogle` (L51-53). If a deployment enables Google but disables email (`NEXT_PUBLIC_EMAIL_LOGIN=false`, `NEXT_PUBLIC_GOOGLE_LOGIN=true`), the modal renders a Google button while the copy still promises "just your email." Make the subtitle conditional on which methods are enabled.

**Nice-to-have**

4. **Mobile bottom sheet still shows a header bar in headerless mode; desktop doesn't.** [UX-heuristic] On mobile, `Modal` routes through `DrawerHeader`, which always renders its `bg-surface-contrast` bar (`apps/dashboard/shared/components/design-system/drawer/drawer-header/DrawerHeader.tsx` L22) — headerless mode only swaps the title for an `sr-only` span, so the bar (and its own close button) stays visible above the login content. Desktop's headerless branch removes the chrome entirely and floats a bare icon over the body instead. That's an inconsistent "headerless" treatment across breakpoints — worth confirming against the Figma mobile frame once access is restored.

5. **Screen-reader accessible name doesn't track the visible step.** [UX-heuristic] The sr-only `DialogPrimitive.Title` is pinned to the static `ariaLabel="Sign in to Anticapture"` (`LoginModal.tsx` L80) across all three views. A screen-reader user always hears "Sign in to Anticapture," even on the "Check your inbox" step, while sighted users see the step-specific heading. Consider passing the current view's heading as the accessible name.

---

- **Keyboard dismiss (Esc) and focus trap**: unaffected by the headerless change — both come from the underlying Radix `Dialog.Root`/`Dialog.Content`, independent of whether a header renders.
- **Accessible name for the headerless dialog**: correctly supplied via a `DialogPrimitive.Title` (sr-only), so Radix's `aria-labelledby` requirement is satisfied — no missing-label a11y warning.
- **Touch targets inside the modal body**: Connect wallet / Email / Google / Resend all use `Button size="lg"` (`py-3`, ≥44px) — fine, only the standalone close `IconButton` (#2) is undersized.

---

### Drafts flow — `ProposalCreationForm.tsx`

Reviewed only for the visible consequence of the session-gating change (must-fix #1 above); the server-derived `isOwner` logic itself is out of scope for a UI pass.

---

**Caveat recap**: Figma MCP could not be authorized in this session — please have a human confirm #3 (copy per config) and #4 (mobile chrome) against the linked frames (node `3315-19892`) before merge. No live app preview was reachable either.

Findings: 5 total — 3 must-fix, 2 nice-to-have.


---

brunod-e:
Consolidado: o conteúdo desta PR foi absorvido pela reorganização em duas PRs — backend #2044 e frontend #2051 (mesmo conteúdo, árvore final idêntica).

PR #2079, opened 2026-07-22
Remove canceled proposals from the activity

pikonha:


isadorable-png:
## 🎨 UI Review

> Automated review · ⚠️ Reviewed without a Figma reference — preview + UX-expert evaluation. No Figma link was found on the PR body/comments, nor on the matching ClickUp tasks (`❌ Don't consider canceled proposals on activity` and `🫕 Panel v2.1` — both have an empty "🕸️ Figma" custom field). Live preview also unreachable: the `anticapture` (dashboard) Vercel deployment was skipped (`Ignored`) for this PR, and the `anticapture-storybook` preview returned 403 (protected, no bypass configured). Findings below are grounded in the diff, the surrounding component code, and standard UX/responsive-layout reasoning.

---

### Panel page — Monitored DAOs table

**The fix itself**
- `wrapperClassName="min-h-[400px]"` → `"min-h-[400px] lg:min-h-0"` (`apps/dashboard/features/panel/components/PanelTable.tsx:227`) removes the 400px height floor at the `lg` breakpoint and up, letting `fillHeight={true}` size the table to the flex space actually available instead of forcing a minimum that was pushing content (and the footer) below the viewport on short viewports. [Code-only]
- This is consistent with the rest of the ancestor chain, which already uses the same `min-h-0`/`flex-1` pattern at `lg:` to make a shrinkable flex column: `PanelSection.tsx` (`lg:min-h-0 lg:flex-1`, `TheSectionLayout` `className="lg:min-h-0"`) down through `SubSectionsContainer` (`lg:min-h-0 lg:flex-1`) and `SubSection`'s `contentClassName="lg:flex lg:flex-col lg:flex-1 lg:min-h-0"`. The `PanelTable` wrapper was the one remaining link in that chain still holding a hard 400px floor — this change makes it match the pattern used everywhere else on this screen. [Code-only]

**Regression check**
- Grepped the repo for other consumers of the same `Table` wrapper pattern: `ServiceProvidersTable.tsx` uses a different `wrapperClassName` (`overflow-x-auto overflow-y-visible`, no `fillHeight`/`min-h` combo) and isn't affected. `PanelTable` is the only place using `fillHeight` + a fixed `min-h-[…]` wrapper, so this change has no blast radius beyond the Monitored DAOs table. [Code-only]

**Breakpoint boundary — nice-to-have**
- The new Playwright test (`apps/dashboard/e2e/panel.spec.ts`) only exercises `1920×640` (a wide desktop width, `lg:` and up, with a short height). It doesn't cover a viewport near the `lg` breakpoint's minimum width with a short height (e.g. ~`1024×640`, a common laptop/split-screen size right at the boundary where the fix's `lg:` class starts applying). Since the whole fix hinges on that breakpoint, consider adding a second viewport case around the `lg` width floor to guard the boundary condition specifically, not just a comfortably-wide desktop width. [UX-heuristic]

**Below `lg` (mobile/tablet) — checked, no issue found**
- Below `lg`, the ancestor chain drops the `h-full`/`flex-1` flex-column layout entirely (`PanelSection`'s outer wrapper is plain `flex flex-col gap-8` there, and `TheSectionLayout`/`SubSection` aren't height-constrained), so the page reverts to normal document flow. A short viewport at these widths simply scrolls the whole page — there's no fixed-height flex container for the table to overflow past, so the footer-overlap bug this PR fixes doesn't reappear at in-between (sub-`lg`) viewport sizes. [Code-only]

---

### Out of scope (noted, not reviewed)
The backend change in `apps/api/src/repositories/proposals-activity/index.ts` (excluding canceled proposals from activity/vote-timestamp queries) has no direct visual surface. It could plausibly change row counts or which rows land in the `isPartiallyIndexed` pinned-to-bottom group in the Monitored DAOs table, but that's a data-correctness question for the code reviewer, not a UI concern — no visual regression expected from it since the table already handles variable row counts.

---

**Summary:** 0 must-fix, 1 nice-to-have (test-coverage), 0 open questions. Small, surgical fix — implementation is consistent with the established layout pattern on this screen.


---

PR #2081, opened 2026-07-22
feat(dashboard): docs links and direct whitelabel SIWE sign-in

brunod-e:
## Summary

Points the product at the new docs site and streamlines whitelabel sign-in: the footer gains an About link (old GitBook) while Docs now goes to https://docs.anticapture.com, the API keys page gets a "See our Docs" button, and whitelabel deployments trigger SIWE directly instead of opening a one-button login modal.

## Context

- 🌿 **Branch**: `feat/docs-links-and-whitelabel-siwe`
- No ticket: requested directly.

## What changed

**UI**
- `Footer.tsx`: link row is now About / Docs / Terms of Service / Give Feedback in both the mobile and desktop blocks. About keeps the old GitBook URL (https://blockful.gitbook.io/anticapture/); Docs points to https://docs.anticapture.com. The mobile row now wraps (`flex-wrap`) since it grew to four links and could overflow narrow phones.
- `ApiKeysManager.tsx`: added a "See our Docs" outline button next to "Create key", linking to https://docs.anticapture.com in a new tab.
- `ApiKeysTable.tsx` (design review feedback): the per-row options menu (a small ellipsis trigger holding a single "Delete key" action) was replaced by a direct, tooltip-labeled destructive trash button that opens the existing delete confirmation modal. The table itself already uses the design-system `Table` component.

**Auth (whitelabel only)**
- `LoginModal.tsx`: whitelabel offers exactly one sign-in method, so opening the login now auto-starts the SIWE flow (wallet picker, then signature) instead of showing a chooser with a single "Connect wallet" button. The dialog only surfaces when there is ceremony state worth seeing: signing/verifying progress or an error with a retry button. Dismissing the wallet picker cancels the sign-in outright (the modal fully closes, so the next click re-triggers cleanly). If the verify timeout resets the flow, the dialog shows the plain connect button as a retry surface.
- Non-whitelabel behavior is unchanged: the auto-start path and the new visibility condition are both gated on `isWhitelabel`, and `handleOpenChange` was only memoized (same body, all deps stable).

## Flow

```mermaid
sequenceDiagram
    participant U as User
    participant H as Header (whitelabel)
    participant P as LoginProvider
    participant M as LoginModal
    participant R as RainbowKit
    participant W as Wallet

    U->>H: click "Connect wallet"
    H->>P: openLogin()
    P->>M: open=true
    Note over M: auto-start (whitelabel): dialog stays hidden
    M->>R: openConnectModal()
    alt user dismisses picker
        R-->>M: closed without connecting
        M->>P: close after grace timer (no stray modal)
    else wallet connects
        R-->>W: connect
        M->>W: SIWE nonce + signMessage
        Note over M: dialog visible: "Check your wallet…"
        W-->>M: signature verified
        P->>P: session lands, close + redirect
    end
```

## Verification

- `pnpm dashboard typecheck` and `pnpm dashboard lint` pass (0 errors; only pre-existing warnings in unrelated files).
- Manually verified on a local dev server (whitelabel via `FORCE_DAO=ENS`): footer links correct in both breakpoint blocks (mobile row wraps at 375px with no horizontal overflow), docs button renders and opens in a new tab with `noopener noreferrer`, whitelabel connect opens RainbowKit directly with no intermediate modal, dismissing leaves no stray dialog, and re-clicking re-triggers the flow.
- All three changes also validated manually by the author locally.

## Self-review summary

- ✅ QA pass: footer breakpoints, external link rel attributes, whitelabel connect/dismiss/retry, non-whitelabel regression check, React 19 StrictMode double-effect check.
- ✅ Engineer pass: no blockers; effect ordering and state resets verified against `LoginProvider`'s wallet/session coherence effects.
- ⚠️ Known limitations (accepted as narrow edge cases):
  - If a wallet approval lands more than 500ms after RainbowKit closes, the pre-existing grace timer cancels the whitelabel sign-in silently (the wallet is then auto-disconnected by the provider's coherence effect). Same 500ms window as before; whitelabel just closes instead of re-showing a chooser.
  - A click while RainbowKit is not yet ready (`openConnectModal` undefined) is a dead click that self-heals via the same grace timer.
  - The API keys header now holds two buttons; on very narrow screens the section title gets tighter.

## Changeset

- `@anticapture/dashboard`: minor, new footer/docs entry points and new whitelabel sign-in behavior.


isadorable-png:
## 🎨 UI Review

> Automated review · [Figma: 🔐 Login & API Keys (partial coverage)](https://www.figma.com/design/mUgy2KpQ3gJ07yZaUaXu8l/%F0%9F%9B%B0%EF%B8%8F-Product-Design?node-id=3315-19892) · [Spec](https://app.clickup.com/t/86ac8e1y5)
> ⚠️ Mixed coverage: the linked spec/Figma covers the API Keys header and the general sign-in modal, but **not** the Footer changes (no ticket — "requested directly" per PR body) or the whitelabel-only sign-in variant (spec explicitly lists "White-label (light) variants" under "still needs design"). Those two areas are reviewed on UX heuristics only.
> ⚠️ The dashboard's own Vercel preview was **Ignored** for this PR (only `anticapture-storybook` deployed) — no live preview was reachable, so regression checks below are diff-level only, not preview-confirmed.

---

### API Keys header (`ApiKeysManager.tsx`)

**Nice-to-have — mobile overflow risk**
- The actions row (`flex shrink-0 items-center gap-2`, line ~122) now holds two buttons ("See our Docs" + "Create key") sitting against `SectionTitle` in a `justify-between` row with no wrap. On narrow viewports (~320–375px) this can squeeze the title/description down to very few characters per line or push the row toward horizontal overflow — the same class of issue this PR's own Footer change had to patch with `flex-wrap`. Suggest `flex-col sm:flex-row` on the outer header row (`apps/dashboard/features/api-keys/ApiKeysManager.tsx:117`), or collapse "See our Docs" to icon-only below `sm` (`hidden sm:inline` on the label). (Self-review already names this as an accepted limitation — flagging in case it's worth fixing rather than accepting, since it isn't a hard tradeoff.) [Code-only]

**Spec drift note**
- The "See our Docs" button doesn't exist in the current Figma reference (node `3375:20590` — API Keys header shows only "+ Create key"). Not a bug, just noting the Figma/ClickUp spec is now behind the implementation — worth a quick sync so the next reviewer isn't confused. [Figma-confirmed]

---

### Footer

No issues found. The new mobile-row `flex-wrap justify-center` (for the 4th link) is a reasonable, low-risk fix. Desktop vs. mobile separators use different existing tokens (`text-dimmed` vs `text-secondary`) but that divergence pre-dates this PR (same pattern already used by the other 3 separators in each row) — not a new inconsistency. [Code-only, no Figma reference exists for this screen]

---

### Whitelabel sign-in (`LoginModal.tsx`)

**Nice-to-have**
- Between the trigger click and RainbowKit's wallet-picker rendering, the sign-in dialog is fully hidden (`whitelabelAutoFlow` suppresses it at idle/connecting) with no loading indicator anywhere. A slow wallet-extension load could read as an unresponsive/dead click. Self-review already flags this as an accepted edge case — consider a lightweight inline spinner near the trigger button (not the modal itself) to close the gap without interfering with the RainbowKit hand-off. [UX-heuristic — no Figma reference exists for the whitelabel variant]

Everything else in this file (auto-start, dismiss-cancels-flow, retry-from-idle) is state logic with no distinct visual surface — out of scope for a UI review.

---

- API Keys header retains the existing `+ Create key` primary button unchanged in position/label — matches Figma (`3375:20590`).
- Sign-in modal's non-whitelabel path (wallet / email / Google chooser) is unchanged by this PR and matches Figma (`3315:19919`).

---

### Scope note


---

PR #2082, opened 2026-07-22
Reduce indexer costs

pikonha:
> [!NOTE]
> **Medium Risk**
> eRPC cache policy changes affect all indexer block fetches and depend on PostgreSQL; misconfiguration could increase latency or upstream spend until alerts fire.
> 
> **Overview**
> **Reduces indexer upstream RPC spend** by tightening eRPC `eth_getBlockByNumber` caching in dev/prod: finalized bodies persist in PostgreSQL (`rpc_cache`), unfinalized/realtime use larger in-memory tiers with chain-aware TTLs, and cache policies no longer apply to all methods. PostgreSQL pool timeouts are set; docs describe `DATABASE_URL` and fall-through behavior.
> 
> **Expands observability** with Postgres and Railway Prometheus exporters (Dockerfiles + Railway configs), new scrape jobs, Grafana panels (eRPC cache, PostgreSQL, Railway RAM/egress), and alerts for cache health, DB pressure, Railway cost signals, plus **15-minute** `keep_firing_for` on existing rules.
> 
> **Dashboard (API keys):** After creating a key, `SaveApiKeyModal` now includes `ConnectAgentSection` with copy-ready MCP install commands while the plaintext is still in memory. `ConnectAgentSection` drops the per-key combobox and auto-picks the newest session key (or placeholder).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 632a3a032cf4dbe0830762fff9518114e19e964d. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


isadorable-png (commented):
## 🎨 UI Review

> Automated review · Figma: not found (searched PR body + ClickUp lists + workspace — no matching task/Figma link for this branch/title) · No ClickUp task linked
> ℹ️ Reviewed without a Figma reference — grounded in code/diff review + UX-expert evaluation.
> ⚠️ Vercel's `anticapture` (dashboard) deployment is **Ignored** on this PR (only `anticapture-storybook` builds) — no live preview available. Findings below are code-level only.

---

### API Keys page — "Connect your AI agent" section

**Must-fix — dropping the key selector removes the only way to see or choose which key is shown**

`ConnectAgentSection` used to let the user pick among their existing API keys via a `Combobox` (`value={selected.id}`, options = every key's label). This PR removes that control entirely (`apps/dashboard/features/api-keys/components/ConnectAgentSection.tsx`) and the component now has **no UI element that names which key is selected** — no combobox, no label, nothing.

On the API Keys page itself (`apps/dashboard/features/api-keys/ApiKeysManager.tsx:183-187`), `ConnectAgentSection` is called with the user's full `keys` list and `lastCreated` (which is `null` on any normal page visit — it's only set for the few seconds right after `handleCreate` succeeds, and resets to `null` on sign-out/sign-in). For any user with **2 or more existing keys** — the ordinary state for a returning user, not an edge case — `selected` falls through to `knownKeys.find(sessionTokens) ?? knownKeys[0]`: an arbitrary/positional pick with no user input and no on-screen indication of which key it is. That user can no longer get the install command for any key other than whichever one happens to land at index 0 (or has a leftover in-session plaintext).

Fix: either keep a lightweight key selector (doesn't need to be the same `Combobox`, but something) for the multi-key case, or at minimum render `selected.label` somewhere in this section so the user knows which key's command/status they're looking at. [Code-only]

**Nice-to-have — client-selector buttons below 44×44px touch target**
`Button size="sm"` (`py-1 px-2`, `text-sm/tight`) is used for the Claude Code / Cursor / Codex selector row (`ConnectAgentSection.tsx:113-121`) — comes in well under the 44px minimum touch target on mobile. Pre-existing (not introduced by this diff), but worth a bump to `size="md"` given this section now also renders inside a modal. [UX-heuristic]

---

### Modal — "Save your API key"

**Validated — no change needed**
`Modal` is given `className="max-w-3xl"` and `bodyClassName="max-h-[70vh] overflow-y-auto"` to fit the now-taller content (`CodeBlock` + `ConnectAgentSection` stacked). Confirmed against `Modal.tsx`: both props are first-class, the mobile drawer conversion is automatic, and `overflow-y-auto` correctly bounds the body on both dialog and drawer variants. [Code-only]

The description copy correctly branches on whether a plaintext token is available (`"Your key is already in it"` vs `"replace the API key placeholder"`) — in this call site `token` is always the freshly created plaintext, so the "already in it" copy is the one that always shows here. Confirmed correct. [Code-only]

### Design-system adherence
No issues — `Modal`, `Button`, `CodeBlock`, `SectionTitle` are all DS components used as intended; no hand-rolled primitives introduced.

### Scope note
This is a UI-only review (visual fidelity, DS adherence, UX, copy, responsive behavior). The rest of this PR (eRPC caching, Postgres/Railway observability, alerting) has no rendered surface and is out of scope here — left to the code reviewer.

---

PR #2086, opened 2026-07-27
feat(dashboard): connect wallet without signing in

brunod-e:
## Problem

Delegating, voting and publishing a proposal are settled on-chain by the wallet's own transaction signature, so they only ever needed a connected account. They were routed through the sign-in modal instead, which forced a SIWE message on anyone who just wanted to act on-chain.

A harder blocker sat in `LoginProvider`: an effect that called `disconnect()` on any connected wallet without a session. Being connected but not signed in was simply not a reachable state.

Applies to both Anticapture and whitelabel.

## What changed

Connecting a wallet and signing in are now separate actions.

| File | Change |
| --- | --- |
| `shared/services/auth/useWalletPrompt.ts` | Always opens the wallet picker, never the sign-in modal |
| `shared/services/auth/LoginProvider.tsx` | Dropped the effect that force-disconnected a session-less wallet; connected without a session is now a valid state |
| `shared/services/auth/LoginProvider.tsx` | Switching accounts still signs a SIWE session out, but no longer re-opens the sign-in modal on its own |
| `shared/components/wallet/ConnectWallet.tsx` | Anticapture header button opens the wallet picker |
| `shared/components/wallet/WhitelabelConnectWallet.tsx` | Same for whitelabel |

No copy changes: the sign-in modal keeps its "Connect wallet" button.

Wallet-born sessions stay bound to their wallet: disconnecting or switching accounts still signs them out, it just never prompts again.

The remaining `openLogin()` call sites are only where a server session is genuinely required: API keys, saving or sharing a draft, and opening the proposal editor. None on the delegate or vote path.

## Verification

`tsc --noEmit` clean and `eslint` at 0 errors, with no warnings in the touched files (the 83 warnings in the app are pre-existing).

Not verified: the wallet flows in a browser. That needs a real wallet extension plus the full local stack, so **please confirm manually that Delegate and Vote open MetaMask without asking for a signature**, on both Anticapture and a whitelabel deployment.

## Worth a look during review

With the header becoming a pure connect affordance, email and Google lose their visible entry point. They now surface only when the user hits a session-gated action (draft, API key). If that is not the intent, the alternative is two separate affordances ("Connect wallet" and "Sign in"), which needs design.


isadorable-png:
## 🎨 UI Review

> Automated review · Figma: [🔐 Login & API Keys — V1 screens](https://www.figma.com/design/mUgy2KpQ3gJ07yZaUaXu8l/%F0%9F%9B%B0%EF%B8%8F-Product-Design?node-id=3315-19892) (found via ClickUp task [`86ac8e1y5`](https://app.clickup.com/t/86ac8e1y5)) · [Spec: 🔐 Login & API Keys](https://app.clickup.com/t/86ac8e1y5)
> ⚠️ Figma link found but the Figma MCP is not connected in this environment (configuration error, not attempted-and-failed) — visual/node-level validation was skipped. Review is grounded in the diff, the linked ClickUp spec text, and UX-expert evaluation.
> ℹ️ Preview: the `anticapture` Vercel deployment was **Ignored** for this PR at review time (only `anticapture-storybook` built) — regression check is diff-level only, not a live before/after comparison.

---

### Header — Connect Wallet button (Anticapture + Whitelabel)

**Question for author — header no longer offers a wallet sign-in path, which the linked spec says it should**
- `ConnectWallet.tsx:70-79` and `WhitelabelConnectWallet.tsx` (same pattern) now open the wallet picker unconditionally, with no SIWE step, in every session state. The linked spec (ClickUp `86ac8e1y5`, "🔐 Login & API Keys") states: *"Wallet auth = SIWE signature under a 'Connect wallet' label; **branch on session state** (logged out = login, logged in = link to current account)."* The Figma sign-in modal (③.1, node `3315:19919`, referenced from that same task) designs wallet-connect as an authentication method, not a connect-only action.
- After this PR, the header's "Connect Wallet" affordance can never authenticate — the only remaining path to sign in with a wallet is the separate sign-in modal, reached only from a session-gated surface (draft, API key). That's exactly the tension flagged in this PR's own "Worth a look during review" section, just without a citation to the spec it diverges from.
- Needs a call before merge: either the spec/Figma get updated to reflect "Connect Wallet" becoming a pure connect affordance (with sign-in moved entirely behind gated actions, as the PR's alternative suggests), or the header should keep the branch-on-session-state behavior the spec calls for. [Code-only — cross-checked against ClickUp spec text, Figma nodes not opened per the caveat above]

---

### Component states

**Nice-to-have — "connected, no session" now looks identical to "signed in" in the header**
- Once a wallet is `connected` (regardless of `session`), `ConnectWallet.tsx` always renders the same avatar/ENS account button (`openAccountModal`). Before this PR, a session-less connected wallet was force-disconnected in `LoginProvider`, so this branch only ever rendered for an actually signed-in user. Now it's reachable while signed out too (e.g. right after voting), and it looks the same. Nothing in the header distinguishes "just connected" from "connected and signed in," so a user could reasonably assume session-gated features (follow/alerts, saved drafts) are available when they aren't.
- Consider a small visual distinguisher on the account button when `!session` — a "not signed in" badge or tooltip — so the two states read differently. (`apps/dashboard/shared/components/wallet/ConnectWallet.tsx:83-108`) [UX-heuristic]

---

### Design-system adherence

No issues — the diff only reconfigures existing `Button` / `ConnectButton.Custom` usage and provider logic; no hand-rolled primitives introduced. [Code-only]

### Mobile

No mobile-specific surface changed in this diff — the same header components render responsively as before; nothing new to check here.

### Scope note

---

PR #2087, opened 2026-07-27
feat: off-chain proposals, voting and visualization (DEV-1131)

brunod-e:
Closes [DEV-1131](https://app.clickup.com/t/90132341641/DEV-1131).

Reworks the off-chain (Snapshot) proposal experience. All 13 final Figma frames
are implemented and mounted. Figma: [🗳️ Off-chain Proposals](https://www.figma.com/design/mUgy2KpQ3gJ07yZaUaXu8l/Product-Design?node-id=3536-25483)

> The ticket points at file key `DEKMQifA8YOb3oxznHboSY`, which only holds a Cover
> page. The frames live in `mUgy2KpQ3gJ07yZaUaXu8l`, the value in the task's
> `🕸️ Figma` field. Worth fixing on the ticket.

## Three root causes worth reading

**1. "Executed" on a Snapshot vote** came from a single line in
`utils/offchainProposal.ts`: `if (for_ > against) return ProposalStatus.EXECUTED`.
Every passing basic ballot read "Executed". Fixed at the source, so the fix lands
on the proposals list and the detail page, not just on new components.

**2. Shuttered elections stuck at 0 / 0.0%** because both indexer syncs are
forward-only on `created`, and a Shutter proposal only reveals its tally *after*
voting closes, by which point the cursors have moved past it. Nothing ever
re-read it. A reveal reconciliation pass now re-fetches closed proposals whose
tally is still all zeros, plus their votes, and upserts without advancing cursors.

**3. Weighted vote weights were in the database the whole time.** The API mapper
ran `Object.keys` over the stored `{choiceIndex: weight}`, keeping indices and
discarding the weights, so a voter's split could not be read back.

## What changed

| Frames | Delivered |
|---|---|
| 01–04, 07, 08 | A designed ballot per vote type. Approval gets an "N of M selected" counter; weighted/quadratic share an allocation ballot with per-option steppers, chart-token dots, stacked bar, `N% Remaining` chip and a total that must reach 100%; ranked gets drag-to-reorder with chevrons as the keyboard path. Lists past 8 options gain a filter and fixed-height scroll. Mobile bottom sheet comes from the design system's `Modal`→`Drawer` |
| 06, 10, 11 | Net-new `OffchainResultsCard`, replacing the ad-hoc results block. Shutter-aware encrypted and reveal-pending states show dashes over empty tracks and never zeros. Optimistic votes render the moment the signature returns, with an indexing chip lifecycle |
| 05, 13 | Read-only "Your Vote" state, entered before the ballot for a wallet that already voted, and the green "You voted" chip on header and mobile bar with a tooltip into the modal |
| 12 | Status derivation mirroring snapshot.box `getProposalState`, quorum-aware, implemented row for row from the frame's table. 16 unit tests, one per row, including one asserting no input can ever produce `executed` |
| 09 | Live impact preview on the single-choice ballot, with the green delta on the selected row |

Allocation colors use the existing `--base-chart-*` tokens, no raw hex. Two DS
gaps the ticket lists were already closed in `main`: the chart tokens and
`Checkbox`. The off-chain non-voters API also already existed.

## Verification

`typecheck` 14/14 tasks, `lint` 0 errors, tests **246/246** dashboard,
**40/40** offchain-indexer, **1005/1005** api. Rebased onto `dev`, no conflicts,
re-verified after.

Every frame was checked in Storybook against its Figma node, comparing computed
styles rather than eyeballing: frame 06 reproduces the design's numbers exactly
(512.4K / 42.1%, 308.9K / 25.4%, …), the Active badge is the white pill on black
text that OQ 5 settled, and the shutter states assert zero zeros. Stories live
under `Governance/` with a Design link per story.

**Not visually verified:** the status badge's fit inside the list's flex row, and
the chip in the header and mobile bar. Both were mounted after their isolated
Storybook check, so spacing in the new context is unconfirmed.

## Decisions that depart from the frames, needing a designer call

- **Frame 09 delta** renders `↑ 26.4K`, the frame shows `↑ 26.4`. Without the
  unit, `26.4` beside `398.2K` reads as 26 tokens rather than 26 thousand
- **Shutter reveal-pending** does not highlight the leading option, though the
  frame shows its `%` in orange. Highlighting a leader while values are dashed
  leaks who is winning. Concealed rows also keep the authored choice order for
  the same reason
- **Frame 09's own numbers** are not internally consistent (3 rows of a longer
  list), so the story uses self-consistent values

## Sequencing and open items

- `weights` only reaches `@anticapture/client` once the API ships and the Gateful
  spec regenerates. Until then the voted modal lists chosen options for weighted
  votes instead of drawing shares. Sequence, not a bug
- **`quorum` / `quorumType` are not exposed.** The quorum branches are
  implemented and tested but inert until the indexer stores those fields
  (new columns + migration) and the API surfaces them
- **OQ 2 still open** (Passed/Rejected vs Approved/Defeated, and the winner copy).
  Labels sit in one `Record`, a one-line change
- **OQ 6** kept as designed: the final card has no quorum row. Nothing was lost,
  the off-chain adapter hardcoded `quorum` to `"0"` so that row never rendered

Also noticed: the repo's `FIGMA_TOKEN` was expired, which silently breaks the
Storybook `addon-designs` links for everyone, and bare `pnpm typecheck` can never
run client codegen because the root script lacks `dotenv --`. Neither is touched
here.


isadorable-png:
## 🎨 UI Review

> Automated review · [Figma: 🗳️ Off-chain Proposals (Voting + Visualization)](https://www.figma.com/design/mUgy2KpQ3gJ07yZaUaXu8l/%F0%9F%9B%B0%EF%B8%8F-Product-Design?node-id=3536-25483) · [Spec (DEV-1131)](https://app.clickup.com/t/90132341641/DEV-1131) · Preview: not reachable
>
> Figma was fetched and cross-checked per frame (all 13 nodes). Vercel previews (app + Storybook) both return 403 (SAML-gated) with no bypass token configured in this environment — review is **Figma + diff only**, no live regression pass. All findings below are grounded in the design context pulled from the Figma nodes listed in the ClickUp task, not from generic heuristics.

---

### Ranked-choice ballot (`RankedChoiceOptions.tsx`, frame 04)

**Must-fix**
- Rank badge doesn't match the frame: Figma's badge is a filled `20×20px` square, `bg-surfaces/contrast (#27272a)`, no border, `Roboto Mono / Medium / 12px`, `text/primary (#fafafa)`. The implementation renders a **bordered** square (`border-border-contrast`) with **no background fill**, `font-inter` instead of mono, and `text-secondary` instead of primary. Fix: swap to `bg-surface-contrast` fill, `font-mono`, `text-primary`, drop the border. (`apps/dashboard/features/governance/components/modals/vote-options/RankedChoiceOptions.tsx:1188`) [Figma-confirmed]

**Nice-to-have**
- Dragged row only picks up the orange `border-highlight` — Figma's "dragging" state also fills the row with `bg-surfaces/contrast (#27272a)` and adds a drop shadow (`0px 4px 12px rgba(0,0,0,0.5)`). Currently the row background stays whatever it was and there's no shadow, so the drag feedback is weaker than designed. (`RankedChoiceOptions.tsx`, `ballotRowClassName`) [Figma-confirmed]

### Single-choice ballot — live impact preview (`SingleChoiceOptions.tsx`, frame 09)

**Nice-to-have**
- Progress-bar track uses `bg-surface-contrast` (`--base-muted`, `#27272a` dark). Figma's in-ballot live-impact bar specifically uses the lighter `surfaces/hover` token (`--base-border`, `#3f3f46` dark) — distinct from the results-card bars, which do use `contrast`. Swap the track to `bg-surface-hover` for this one component. (`SingleChoiceOptions.tsx`) [Figma-confirmed, verified against `apps/dashboard/app/globals.css` token values, not just the fallback hex]

**Question for author / Isadora**
- Delta shows `↑ 26.4K`; the frame shows `↑ 26.4` with no unit. Already flagged in the PR description as a deliberate departure — raising it here so it gets an explicit yes/no rather than living only in the PR body. Worth keeping the `K` for legibility (bare "26.4" next to "398.2K" does read as 26 tokens), but that's a call for Isadora, not an engineering default. [Figma-confirmed]

### Voted-state modal (`OffchainVotedModal.tsx`, frame 05)

**Nice-to-have**
- Unchosen (0%) options render an empty `bg-surface-contrast` bar track under the row. Figma's 0% row (`Grants program`) has no bar track element at all for that state — just the row header. Minor, but worth dropping the track entirely when `percent === 0` to match. [Figma-confirmed]

### Shutter-encrypted results (`OffchainResultsCard.tsx`, frame 11)

**Question for author / Isadora**
- Also already flagged in the PR: no leading-option highlight while concealed, and authored order is kept instead of sorting. That is the right call for not leaking the winner — but the raw Figma frame (both "voting open" and "reveal pending" states) still renders the **first row's percent value in `text/link` orange** even though the values themselves are dashed. That's either (a) an intentional "always highlight row 1 regardless of concealment" instruction, or (b) a leftover artifact from copy-pasting the results-card frame (06) as the base for frame 11 and not stripping the orange styling. Worth a quick confirm with Isadora on which it is, so the Figma file gets cleaned up either way (currently ambiguous evidence for future readers of the frame). [Figma-confirmed]

### Status badges (`OffchainProposalBadge.tsx`, frame 12) — validated

- Pending/Active/Passed/Rejected/Closed variants (`dimmed` / `primary` w/ white-pill-black-text override / `success` / `error` / `outline`) match the frame's badge specimens exactly, including the Active-is-a-white-pill decision from OQ 5.
- `winner: X · NN%` in `font-mono text-xs text-secondary` matches the frame's Roboto Mono / 12px / secondary treatment.
- Reminder: **OQ 2** (Passed/Rejected vs Approved/Defeated label vocabulary + winner copy wording) is still explicitly open on the ClickUp task, owned by Alex — not blocking, just flagging so it doesn't get lost now that the derivation logic itself has shipped.
- Reminder: **OQ 6** (no quorum row on the final results card) is owned jointly by Isadora/Alex per the ticket. The PR addresses it head-on ("kept as designed... nothing was lost, quorum was hardcoded to 0") — surfacing here only so the ClickUp OQ gets formally closed rather than left open with the behavior already shipped.

### Has-voted indicator (`OffchainVotedChip.tsx` / `ProposalHeader.tsx` / `ProposalSection.tsx`, frame 13)

**Must-fix**
- The green "✓ You voted For" chip is wired into the proposal header and the mobile bottom bar, but **not into the proposals list** (`ProposalItem.tsx`). Frame 13 is explicitly labeled "PROPOSALS LIST" in Figma and shows the chip inline in a list row's detail line, and the ClickUp AC says the chip must be "visible on header + list." The PR description's own summary table narrows this to "header and mobile bar" without calling out the list as a scope cut — worth confirming whether that's intentional or just missed, since as written it doesn't meet the AC. [Figma-confirmed] [ClickUp-AC-confirmed]

### Weighted / Quadratic ballot (`AllocationVoteOptions.tsx`, frame 03) — validated

- Allocation dot + bar-segment colors resolve to `--base-chart-1` (`#3b82f6`), `-2` (`#ec4899`), `-3` (`#f59e0b`) — exact match to the frame's raw hex, and the DS gap the ticket called out (chart tokens, no raw hex) is genuinely closed.
- Stepper sizing (26px buttons, 48px value cell), "20% Remaining" badge, "Total: 80%" amber-when-incomplete / green-when-complete, and the 2px gap between bar segments all match the frame.

### Approval ballot (`ApprovalVoteOptions.tsx`, frame 02) / Overflow (`BallotSection.tsx`, frame 07) — validated

- Selected-row orange border + filled checkbox, "N of M selected" counter placement, filter input + fixed-height scroll + "Scroll to see all N options" helper all match their frames structurally and by copy.

---

### Mobile

Not independently re-verified against frame 08 (mobile bottom sheet) or a live preview — the PR reuses the existing DS `Modal`→`Drawer` conversion, which is the correct pattern per the spec, but the "not visually verified" caveat the author already flagged for the header/mobile-bar chip fit stands; worth a quick pass once the preview is reachable.

---

- Status badge palette and Active-pill treatment (frame 12).
- Allocation colors (chart tokens, frame 03).
- Approval ballot selection styling and counter (frame 02).
- Overflow filter/scroll pattern (frame 07).
- `winner: X · NN%` mono styling (frames 11/12).
- "You voted For" badge variant/icon (frame 13).

---

**Summary:** 2 must-fix, 3 nice-to-have, 3 open questions for Isadora/the author (all tied to decisions the PR itself flagged as needing a designer call, plus the OQ 2/OQ 6 reminders already sitting on the ClickUp task).

---

pikonha (approved):

PR #2088, opened 2026-07-27
Dev

pikonha:
> [!NOTE]
> **Medium Risk**
> Touches auth credential boundaries (usage-only vs provisioning keys), new DB migrations in Authful, and wallet RPC routing—misconfiguration could break sign-in or RPC until env vars are set correctly.
> 
> **Overview**
> Adds an end-to-end **30-day daily request chart** for self-service API keys: **Gateful** buffers `user:*` traffic and flushes idempotent batches to **Authful** via a new **usage-only** credential (`USAGE_API_KEY` / `TOKEN_SERVICE_USAGE_API_KEY`) that cannot mint or revoke; **Authful** persists `token_usage_daily` with tenant-scoped read/write rules; **User API** exposes `GET /me/api-keys/usage`; the **dashboard** renders a filterable stacked bar chart and related API Keys UX (docs link, per-row delete, agent commands in the save modal).
> 
> **Dashboard wallet RPC** no longer uses `NEXT_PUBLIC_ALCHEMY_KEY`—wagmi routes through a same-origin **`/api/rpc/<chainId>`** proxy that injects `ERPC_SECRET` server-side and forwards trusted client IP for per-IP limits.
> 
> **DAO API** proposal-activity queries now **exclude canceled proposals** (including first-vote timestamp alignment). **Indexer** corrects ENS token `startBlock`. **eRPC** configs gain dashboard integration docs, trusted forwarders, failsafe/rate-limit/cache tuning, and Chainstack fallback tagging; **monitoring** adds Postgres/Railway exporter docs and alert guidance. Minor **panel** layout and **whitelabel** sign-in (direct SIWE) fixes; **AGENTS.md** documents ClickUp issue tracker and agent skills.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit db90edd16d10b5337be78fbb54f81027613e8cfe. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #2089, opened 2026-07-27
Chore/prod deploy fixes

pikonha:
> [!NOTE]
> **High Risk**
> Touches internet-facing auth (new usage credential and Gateful flush/shutdown), wallet RPC routing, and production eRPC rate limits/caching—misconfiguration could break sign-in, key usage visibility, or wallet connectivity.
> 
> **Overview**
> Adds an end-to-end **30-day daily request chart** for self-service API keys: **Gateful** buffers `user:*` traffic and flushes idempotent batches to **Authful** on a new **usage-only** credential (`USAGE_API_KEY` / `TOKEN_SERVICE_USAGE_API_KEY`) that cannot mint or revoke; **Authful** stores `token_usage_daily` with batch dedup and tenant-scoped reads; **User API** exposes `GET /me/api-keys/usage`; the **dashboard** renders a stacked bar chart with per-key filtering and polishes the API Keys UX (docs link, direct delete, agent commands in the create modal).
> 
> **Wallet RPC** no longer uses `NEXT_PUBLIC_ALCHEMY_KEY` in the browser—wagmi goes through a server **`/api/rpc/[chainId]`** proxy that injects `ERPC_SECRET` and forwards a validated client IP for per-IP limits. **eRPC** configs gain trusted forwarders, tuned failsafe/hedging, Chainstack as `tier:fallback`, higher budgets, and **`eth_getBlockByNumber`** block caching (Postgres + in-memory tiers). Docs and **`.env.example`** switch required dashboard RPC vars to `ERPC_URL` / `ERPC_SECRET`.
> 
> Smaller deploy fixes: **DAO API** proposal activity excludes **CANCELED** proposals (and votes on them); **indexer** corrects **ENS** token `startBlock`; panel table height overflow; footer links; whitelabel sign-in starts **SIWE** directly; Tornado Cash icon update; **monitoring** Dockerfiles for Postgres and Railway exporters; **AGENTS.md** ClickUp/triage/domain agent docs.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit aa64def2b5eb3eea31ee45cf002f290b2e09066e. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


isadorable-png:
## 🎨 UI Review

> Automated review · Figma: [🔐 Login & API Keys — V1 screens](https://www.figma.com/design/mUgy2KpQ3gJ07yZaUaXu8l/%F0%9F%9B%B0%EF%B8%8F-Product-Design?node-id=3315-19892) (found via ClickUp task [`86ac8e1y5`](https://app.clickup.com/t/86ac8e1y5)) · [Spec: 🔐 Login & API Keys](https://app.clickup.com/t/86ac8e1y5)
> ℹ️ This is a small follow-up PR that implements the fixes recommended in the automated review on PR #2088 (same branch lineage). All three main fixes were verified directly against the design-system source (`Button`, `Select`, chart tokens) rather than against Figma, because the touched surfaces are mobile-only or logic-only, and this Figma file only documents desktop frames (1280×720).
> ⚠️ Preview: only `anticapture-storybook` deployed for this PR; the `anticapture` dashboard deployment itself was **Ignored** — no live before/after comparison, this review is diff + source-level verification only.

---

### API Keys page — Row actions (`ApiKeysTable.tsx`)

**Fixed — mobile touch target restored**
- `min-h-11 min-w-11 lg:min-h-0 lg:min-w-0` is back on the delete button (`ApiKeysTable.tsx:32`). Verified against `Button.tsx`: `size="md"` only sets `py-2 px-4` with `h-fit` (no fixed height), so without the override the button would render under 44px on mobile. Nothing else in the class list conflicts with the added `min-h-*`/`min-w-*` utilities, so the 44×44px target is genuinely restored below `lg`, and the button shrinks back to compact at `lg` as before. [Code-only] — verified against DS source, closes the regression flagged on PR #2088.

### API Keys page — Usage chart (`UsageSection.tsx`)

**Fixed — DS `Select` replaces the native `<select>`**
- The mobile key filter now uses `<Select items={filterOptions} value={activeKeyId} onValueChange={setSelectedKeyId} aria-label="Filter usage by API key" className="w-36 lg:hidden" />`. Checked `Select`'s prop type (`SelectItem = { value, label }`) against `filterOptions` (`{ label, value: id }[]`) — the shapes line up exactly, so this is a correct swap, not just a visual patch. [Code-only] — verified against DS source, closes the finding from PR #2088.

**Fixed — chart series colors now resolve from design tokens**
- `getChartSeriesColors()` (`chart-theme.ts`) reads `--base-chart-1` through `--base-chart-7` via `getComputedStyle`, falling back to the same 7 hex values only when `window` is undefined (SSR). Checked `globals.css`: `--base-chart-1..7` are defined identically in both the default and `.dark` blocks (`#3b82f6`, `#ec4899`, `#f59e0b`, `#a855f7`, `#10b981`, `#06b6d4`, `#fbbf24`) — so unlike the token-fallback trap that bit the `ProgressCircle` ring in PR #2084, there's no light/dark divergence to worry about here. `transform.ts` now sources the palette from this helper instead of its own hardcoded array. [Code-only] — verified against DS source and `globals.css`, closes the finding from PR #2088.

### Sign-in — Whitelabel direct SIWE (`LoginModal.tsx`)

**Fixed (mostly) — silent auto-close on a picker that never opened**
- `pickerSeen` now gates the whitelabel auto-close: the sign-in modal only self-dismisses if RainbowKit's wallet picker actually opened during this session (`connectModalOpen` went true at least once). Previously it closed after 500ms even when the picker never appeared, which read as "nothing happened" — the question raised on PR #2086/#2088. Now, when the picker never opens, `siweReset()` still fires but the modal is **not** forced closed — it reappears with the manual "Connect wallet" button once `whitelabelAutoFlow` re-evaluates to false.
- **Still worth a look — brief fully-blank window:** for that same ~500ms while `siwe.status === "connecting"` and the picker hasn't opened, `whitelabelAutoFlow` is true, which hides the `LoginModal` itself (`open={open && !connectModalOpen && !whitelabelAutoFlow}`), and RainbowKit's own dialog isn't up either — so there's a brief moment with **no visible dialog at all** before the "Connect wallet" button reappears. This isn't new (the same hidden-during-`connecting` state existed before this PR) and it self-resolves in ~500ms, but on a slow connector init it could still read as a flicker/hang rather than a loading state. Worth a lightweight loading indicator on the trigger for that window if it comes up in QA — not a blocker. [UX-heuristic] — nice-to-have, not a regression from this diff.

---

### Icons

**No issues — Tornado Cash icon swap**
- `TornadoCashIcon.tsx` now uses the icon's actual multi-path artwork (mint `#45F0A9` badge, black glyph) instead of the previous hand-drawn placeholder swirl. Hardcoded brand colors here are expected — DAO/protocol logos aren't meant to route through semantic tokens. [Code-only]

---

### Design-system adherence

No new violations — this PR replaces two prior DS violations (hand-rolled `<select>`, hardcoded chart palette) with their token-driven / DS-component equivalents; nothing hand-rolled was introduced.

### Mobile

Covered above — both mobile-specific fixes (touch target, filter `Select`) are correctly implemented per source-level verification.

### Scope note

---

PR #2094, opened 2026-07-29
chore: trigger client build

pikonha:
> [!NOTE]
> **Low Risk**
> Import-only reorder in codegen config; no logic, API, or security impact.
> 
> **Overview**
> Reorders the `./src/generators` import list in `kubb.config.ts` so `mapEthereumFormatFakers` is imported after `mapEthereumFormatTypes` instead of before it. **No runtime or codegen behavior changes**—same symbols, same Kubb plugin wiring.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit fecea54447efa4a5b01541358c6b25f1ef140867. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/apps/dashboard/e2e/governance.spec.ts
+++ b/apps/dashboard/e2e/governance.spec.ts
@@ -84,6 +84,59 @@ test.describe("Governance page (/ens/proposals)", () => {
     await expect(proposalLink.locator("h3")).not.toHaveText("");
   });
 
+  test("shows Snapshot quorum identifier on proposal progress bar", async ({
+    goto,
+    page,
+  }) => {
+    await page.route(
+      /\/api\/gateful\/ens\/offchain\/proposals(\?.*)?$/,
+      async (route) => {
+        await route.fulfill({
+          contentType: "application/json",
+          body: JSON.stringify({
+            items: [
+              {
+                id: "snapshot-quorum-marker",
+                spaceId: "ens.eth",
+                author: "0x0000000000000000000000000000000000000001",
+                title: "Snapshot quorum marker proposal",
+                body: "",
+                discussion: "",
+                type: "basic",
+                start: 1767225600,
+                end: 1798761600,
+                state: "active",
+                created: 1767225600,
+                updated: 1767225600,
+                link: "https://snapshot.box/#/s:ens.eth/proposal/snapshot-quorum-marker",
+                flagged: false,
+                scores: [60, 20, 10],
+                scoresTotal: 90,
+                quorum: 45,
+                choices: ["For", "Against", "Abstain"],
+                network: "1",
+                snapshot: null,
+                strategies: [],
+              },
+            ],
+            totalCount: 1,
+          }),
+        });
+      },
+    );
+
+    await goto("/ens/proposals?source=snapshot");
+
+    const proposalLink = await waitForProposalLink(page);
+    await expect(proposalLink).toContainText("Snapshot quorum marker proposal");
+    await expect(
+      proposalLink.getByLabel("Snapshot quorum marker"),
+    ).toBeVisible();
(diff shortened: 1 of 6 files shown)


isadorable-png:
## 🎨 UI Review

> Automated review · Figma: found but scope-mismatched — [ClickUp DEV task "🗳️ Off-chain Proposals (Voting + Visualization)"](https://app.clickup.com/t/86aje4uz4) links a Figma file for the Snapshot voting/ballot redesign, but that file has no frame depicting the proposal-list quorum marker or the proposal-not-found blank slate — the two things this PR actually changes. Findings below are diff + UX-expert evaluation, not Figma-validated.
> ⚠️ This PR's title/description ("chore: trigger client build," import-reorder only) is stale — a second commit (`fix: add proposal not found state`) landed after the Cursor Bugbot summary was written and is the real substance of this diff. Flagging in case the description needs updating before merge.
> Live preview (`anticapture-ei6lklo1t-ful.vercel.app`) was not rendered in this pass — no headless browser available in this environment — so sizing/spacing claims below are read from the diff, not pixel-confirmed.

**Scope:** `apps/dashboard/features/governance/components/proposal-overview/` (proposal list quorum marker + proposal-detail not-found state). The `kubb.config.ts` import reorder is a no-op, out of scope.

---

### Governance proposal list — vote progress bar

**Must-fix — raw error text shown to users**
`ProposalSection.tsx`'s new fallback for non-404 errors renders `error.message` directly as the BlankSlate description:
```tsx
<BlankSlate title="Unable to load proposal" description={error.message} .../>
```
`error.message` comes straight from the fetch/query client (e.g. a raw `Not Found`, a network failure string, or whatever the generated client surfaces) — it's not written for end users and can leak implementation detail. Use a fixed, user-facing copy string here (e.g. "Something went wrong loading this proposal. Try refreshing.") instead of the raw error. (`apps/dashboard/features/governance/components/proposal-overview/ProposalSection.tsx`)

**Nice-to-have — bar/marker size changed for every proposal, not just the new Snapshot ones**
The refactor into a shared `ProposalQuorumProgress` also changed the visual spec for both on-chain and off-chain cards: bar track `h-1` → `h-1.5`, quorum marker `h-2 w-[2px]` → `h-3.5 w-[3px]`. That's a size bump on every proposal card in the list, not just the ones that gained the new Snapshot marker. With no Figma frame to check this against, worth a quick design confirmation that the new size is intentional rather than an incidental side effect of the refactor.

**Nice-to-have — same icon/tone for "not found" vs "failed to load"**
Both the not-found case and the generic-error case use `icon={AlertOctagon}` with `variant="default"`. Consider differentiating (e.g. a neutral/search icon for "doesn't exist" vs. the alert icon reserved for the true error) so users can tell "this proposal was never here" apart from "something broke, try again." (`ProposalSection.tsx`)

- `isProposalNotFoundError` correctly narrows both the direct `.status` and nested `.response.status` 404 shapes, and is unit-tested (`proposalErrors.test.ts`) for both cases plus a non-404 negative case.
- The not-found path now renders the DS `BlankSlate` component instead of the old raw `<div className="text-primary p-4">Proposal not found</div>` — correct DS reuse, no hand-rolled error markup.
- New e2e coverage (`governance.spec.ts`) confirms the blank slate renders (and the generic `route-error-fallback` does not) for a 404'd proposal, and separately confirms the Snapshot quorum marker + "Quorum: 45" tooltip render for an offchain proposal — good regression coverage for exactly the two things this PR changes.

### Mobile
No mobile-specific code in this diff; the progress bar and blank slate inherit existing responsive behavior from their parents — nothing new to check here.

### Scope note
This is a UI-only review (visual fidelity, DS adherence, UX, copy) — architecture and hook/test structure are out of scope; the author has a code reviewer for that.

---

PR #2101, opened 2026-08-01
fix(dashboard): remove 10k character limit on proposal description

alextnetto:
## Summary

Hotfix: the create-proposal form blocked proposal descriptions over 10,000 characters, preventing long governance proposals from being published. The limit is removed — the description now has no client-side length cap.

## Changes

- `schema.ts` — dropped `.max(10_000)` from the `body` field validation
- `ProposalCreationForm.tsx` — removed the `<= 10_000` condition from the `canPublish` gate
- `BodyField.tsx` — replaced the `x / 10,000` limit counter (with warning/error colors) with a plain character count
- `constants.ts` — removed the now-unused `BODY_CHAR_LIMIT` and `BODY_WARNING_THRESHOLD`
- `schema.test.ts` — flipped the limit test to assert bodies over 10k are accepted

## Note for reviewers

The user-api drafts endpoint still caps `body` at 100,000 characters (`BODY_MAX` in `apps/user-api/src/mappers/drafts/index.ts`) as a payload-size safeguard on the drafts store. That cap was left in place — it only affects saving drafts, not publishing, and removing a server-side abuse guard felt out of scope for this hotfix. Flag if it should go too.

## Verification

- `pnpm dashboard typecheck` ✅
- `pnpm dashboard lint` ✅
- `pnpm dashboard test` — 44 suites, 266 tests, all passing ✅

### DIFF
--- a/.changeset/proposal-body-no-limit.md
+++ b/.changeset/proposal-body-no-limit.md
@@ -2,4 +2,4 @@
 "@anticapture/dashboard": patch
 ---
 
-Remove the 10,000-character limit on the proposal description in the create-proposal form. The body field now accepts descriptions of any length, and the editor footer shows a plain character count instead of a limit counter.
+Raise the proposal description limit in the create-proposal form from 10,000 to 100,000 characters, matching the ceiling the drafts endpoint already enforces. Long governance proposals are no longer blocked from being published, and the editor footer counter warns as the new limit approaches instead of failing with a generic error.
--- a/.changeset/proposal-body-no-limit.md
+++ b/.changeset/proposal-body-no-limit.md
@@ -1,0 +1,5 @@
+---
+"@anticapture/dashboard": patch
+---
+
+Remove the 10,000-character limit on the proposal description in the create-proposal form. The body field now accepts descriptions of any length, and the editor footer shows a plain character count instead of a limit counter.
--- a/apps/dashboard/features/create-proposal/components/BodyField.tsx
+++ b/apps/dashboard/features/create-proposal/components/BodyField.tsx
@@ -44,6 +44,10 @@ const markdownHighlightStyle = HighlightStyle.define([
 const sourceViewExtensions = [syntaxHighlighting(markdownHighlightStyle)];
 
 import { TabGroup } from "@/shared/components/design-system/tabs/tab-group/TabGroup";
+import {
+  BODY_CHAR_LIMIT,
+  BODY_WARNING_THRESHOLD,
+} from "@/features/create-proposal/constants";
 import type { ProposalFormValues } from "@/features/create-proposal/schema";
 
 type Mode = "visual" | "markdown";
@@ -56,6 +60,13 @@ export const BodyField = ({ version = 0 }: BodyFieldProps) => {
   const { control, watch } = useFormContext<ProposalFormValues>();
   const body = watch("body") ?? "";
   const [mode, setMode] = useState<Mode>("visual");
+
+  const counterColor =
+    body.length > BODY_CHAR_LIMIT
+      ? "text-error"
+      : body.length >= BODY_WARNING_THRESHOLD
+        ? "text-warning"
+        : "text-secondary";
 
   return (
     <div className="flex w-full flex-col gap-1">
@@ -154,8 +165,8 @@ export const BodyField = ({ version = 0 }: BodyFieldProps) => {
           )}
         />
       </div>
-      <p className="text-secondary text-xs">
-        {body.length.toLocaleString()} characters
+      <p className={`${counterColor} text-xs`}>
+        {body.length.toLocaleString()} / {BODY_CHAR_LIMIT.toLocaleString()}
       </p>
     </div>
   );
--- a/apps/dashboard/features/create-proposal/components/BodyField.tsx
+++ b/apps/dashboard/features/create-proposal/components/BodyField.tsx
@@ -44,10 +44,6 @@ const markdownHighlightStyle = HighlightStyle.define([
 const sourceViewExtensions = [syntaxHighlighting(markdownHighlightStyle)];
 
 import { TabGroup } from "@/shared/components/design-system/tabs/tab-group/TabGroup";
-import {
-  BODY_CHAR_LIMIT,
-  BODY_WARNING_THRESHOLD,
-} from "@/features/create-proposal/constants";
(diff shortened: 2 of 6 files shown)


isadorable-png:
## 🎨 UI Review

> Automated review · Figma: link found (`mUgy2KpQ3gJ07yZaUaXu8l`, node `1965:70112`) but the Figma MCP in this environment only sees a locally-open "Cover" page — target nodes return "not found," so this is a configuration error, not a missing reference · Spec: [📬 Create proposal](https://app.clickup.com/t/86agjwww8) (ClickUp, released)
> ⚠️ Figma link found but Figma MCP unreachable — visual validation skipped, configuration error reported. Grounded in the linked ClickUp spec (a strong, explicit text source) + diff review + UX-expert evaluation.
> ℹ️ No live preview either: the `anticapture` dashboard Vercel deployment for this PR shows as **Ignored** (only the unrelated `anticapture-storybook` project builds) — consistent with other recent PRs in this repo, not specific to this change. Findings below are `[ClickUp-confirmed]` or `[Code-only]`.

---

### Create Proposal form — Body field

**Question for author — this hotfix reverts a documented, acceptance-criteria-level product decision** — `[ClickUp-confirmed]`
The linked spec (ClickUp `📬 Create proposal`, status "released") is explicit and specific:
> Form Fields table: "Body | Rich text editor | Required; **10,000 character limit**; live counter **'0 / 10,000'**"
> Acceptance Criteria: "The Body field **enforces a 10,000-character limit** with a live counter; **publish is blocked at the limit**"

This PR removes exactly that: the `.max(10_000)` schema validation, the `canPublish` gate condition, and the color-coded limit counter (`X / 10,000`, warning past 9,500, error past 10,000) — replacing it with a plain, unlimited `X characters` count (`BodyField.tsx:154-161`, `schema.ts:73`, `ProposalCreationForm.tsx:263-268`). The PR description frames this purely as a bug fix ("blocked...long governance proposals from being published"), but the spec treats the 10k cap as an intentional, shipped decision, not a bug. Before merging: has product/design signed off on lifting the cap entirely, or should there be a *new*, higher limit rather than none? If the cap is being intentionally retired, the ClickUp spec should be updated so it doesn't misrepresent the shipped behavior to the next reader.

**Nice-to-have — the counter now gives zero feedback at any length** — `[Code-only]`
Even if the hard cap is being retired, this change also deleted all conditional styling (`counterColor`: `text-secondary` → `text-warning` → `text-error`), not just the block. The counter is now a flat, uncolored `X characters` string (`BodyField.tsx:157-159`) with no signal at any length. Meanwhile `user-api`'s drafts endpoint still caps stored bodies at 100,000 chars (per the PR's own note), and the on-chain publish path has no documented size ceiling at all now. A user who writes a 50,000+ character proposal gets no visual cue they're in unusual territory until something fails downstream (a revert, an oversized calldata payload, a slow render) — worse UX than the removed cap, which at least gave upfront, actionable feedback. Consider keeping a soft warning threshold (e.g. re-introduce `text-warning` past some size) even without a hard block.

---

### Mobile

No layout or responsive-relevant code in this diff (only a validation rule and a counter's color logic changed) — nothing new to verify here.

---

### Design-system adherence

No issues — the diff only removes conditional Tailwind classes on an existing `<p>`; no hand-rolled primitives introduced.

### Scope note
This is a UI-only review (copy/product-fidelity vs. the linked spec, DS adherence, UX) — architecture, validation-layer design, and other code-quality concerns are left to the code reviewer.

---

alextnetto:
Addressed in e0fc7a608 — the Codex point was right, and it turned out to be the answer to the UI reviewer's counter-feedback point too.

## 1. Body validation vs. draft storage limit (Codex, P2)

Confirmed and fixed. Dropping the cap entirely let the form accept bodies the drafts endpoint rejects — `BODY_MAX = 100_000` in `apps/user-api/src/mappers/drafts/index.ts:51`, enforced on both `CreateDraftBodySchema` and `UpdateDraftBodySchema`. The form now enforces the same ceiling rather than no ceiling:

- `constants.ts` — `BODY_CHAR_LIMIT = 100_000` restored, with a comment pointing at the user-api constant it mirrors so the two do not drift.
- `schema.ts` — `.max(BODY_CHAR_LIMIT, "100,000 character limit")`, so the body field shows an inline error.
- `ProposalCreationForm.tsx` — worth flagging that the schema alone was **not** enough here: neither `handleSaveDraft` nor `handleShare` runs schema validation before POSTing, they read `form.watch()` values directly. Both now short-circuit with an explicit "Description is over the 100,000 character limit" toast instead of the generic "Could not save draft".

## 2. Counter gives zero feedback (UI review, nice-to-have)

Resolved by the same change rather than as a separate soft warning. The counter is back to `X / 100,000` with `BODY_WARNING_THRESHOLD = 95_000` driving `text-secondary` → `text-warning` → `text-error`. Same upfront, actionable feedback as before — just at the ceiling that actually exists downstream instead of an arbitrary one.

## 3. Reverting a documented product decision (UI review, question) — needs product sign-off

This one I can't resolve in code, so flagging it rather than closing it. The reviewer is right that the ClickUp spec [📬 Create proposal](https://app.clickup.com/t/86agjwww8) lists the 10k cap and the `0 / 10,000` counter as acceptance criteria, so this is a deliberate shipped decision being changed, not a bug being fixed.

What this PR now does is narrower than what it originally did — it raises the cap 10k → 100k rather than removing it, and 100k is not arbitrary: it is the limit the persistence layer already enforces. That should be an easier call than "no limit at all". But it is still a product change, so:

- **Someone on product should confirm 100k is the intended new cap**, and
- **the ClickUp spec needs updating** — its Form Fields table and Acceptance Criteria still say 10,000 and will misrepresent shipped behaviour otherwise. I have not edited the spec; happy to if that is wanted.

If product would rather keep a cap below 100k, it is a one-line change to `BODY_CHAR_LIMIT`.

## Verification

- `tsc --noEmit` on the dashboard ✅
- ESLint + Prettier on `features/create-proposal` ✅
- `pnpm dashboard test` — 44 suites, **268** tests, all passing ✅ (up from 266: added cases asserting 100,000 is accepted and 100,001 rejected, alongside the existing over-10k-accepted case)

Not rendered in a browser — the `anticapture` Vercel deployment shows as Ignored on this PR, same as the reviewer observed. The counter colour change is a conditional class on an existing element.

pikonha (approved):

PR #2138, opened 2026-08-19
chore(dashboard): comment the Shutter DAO out of the DAO list

brunod-e:
Comments `SHU` out of `DaoIdEnum` (same move as the commented-out `OPTIMISM`), which removes Shutter from `ALL_DAOS` and everywhere the app enumerates DAOs: the panel table, routing (`/shu` now 404s), the shutter.gov.blockful.io hostname mapping, sitemap/OG surfaces.

## Fallout handled

Unlike OP, Shutter had live `DaoIdEnum.SHU` references, disabled minimally and revertibly:

- `create-proposal/constants.ts` — SHU line commented out of `canCreateProposalForDao`.
- `submitProposalRequest.ts` — `isAzoriusDao` returns `false` (SHU was the only Azorius DAO); the Azorius submit path stays in place, just unreachable.
- `voteOnProposal.ts` / `submitGovernanceAction.ts` — the Azorius vote handler, execute ABI, and their `case DaoIdEnum.SHU` branches are commented out, so uncommenting restores them verbatim.
- `ProposalHeader.tsx` / `ProposalSection.tsx` — the `isShu` checks compare against the raw `"SHU"` string (they take `daoId: string`), keeping the queue/execute logic intact.
- `useCalldataReview.ts` — SHU entry commented out of the dao-proposals repo map.
- `DaoOverviewHeader.tsx` — drops the SHU-only 4-decimal price formatting.

`shared/dao-config/shu.ts` and its registration stay, matching how `OP` is kept in `rawDaoConfigByDaoId` with its enum member commented.

## Verification

`tsc --noEmit` clean · `eslint` 0 errors · 527 tests / 54 suites pass.


isadorable-png:
## 🎨 UI Review

> Automated review · Figma: not found (PR body/comments, ClickUp lists 901320776629/901323162483, and workspace search all came up empty; the matching ClickUp ticket [DEV-1155 "Disable Shutter entirely in production"](https://app.clickup.com/t/86ak1jkc3) has no description/attachments) · Preview: unreachable from this review environment (network egress blocked to the Vercel preview domain) — regression check below is diff + source-level only, no live preview.
> ℹ️ Reviewed without a Figma reference or live preview — grounded in source-level tracing of the DAO-enumeration paths this PR's enum change ripples into, plus UX-expert judgment on the touched components.

This is a narrow, well-executed change — the `DaoIdEnum.SHU` removal is applied consistently everywhere it's referenced directly (`create-proposal/constants.ts`, `submitProposalRequest.ts`, `useCalldataReview.ts`, `submitGovernanceAction.ts`, `voteOnProposal.ts`), all as commented-out/revertible code matching the existing `OP` precedent, and `ProposalHeader.tsx`/`ProposalSection.tsx` correctly preserve their queue/execute logic by comparing against the raw `"SHU"` string. One gap surfaced by tracing where else the app enumerates DAOs.

---

### Sitemap & `llms.txt` — dead `/shu/*` links survive the disablement

The PR description states this change removes Shutter from "everywhere the app enumerates DAOs: the panel table, routing (`/shu` now 404s), the `shutter.gov.blockful.io` hostname mapping, sitemap/OG surfaces." The panel table, hostname mapping, and per-route 404 behavior all check out — but two public, crawlable surfaces don't:

- **`apps/dashboard/app/sitemap.ts:135`** — `const daoIds = Object.keys(daoConfigByDaoId).map((id) => id.toLowerCase());`
- **`apps/dashboard/app/llms.txt/route.ts:8`** — `const daoUrls = Object.keys(daoConfigByDaoId).map(...)`

Both iterate the keys of `daoConfigByDaoId` directly, not `ALL_DAOS` (`= Object.values(DaoIdEnum)`). Since `rawDaoConfigByDaoId` in `shared/dao-config/index.ts` still has `SHU` as a plain object-shorthand key (independent of the enum), `Object.keys(daoConfigByDaoId)` still includes `"SHU"` even after this PR. Concretely:

- `sitemap.ts`'s `daoEntries` will keep emitting `/shu`, `/shu/proposals`, `/shu/risk-analysis`, `/shu/token-distribution`, `/shu/stakeholders`, `/shu/activity-feed`, `/shu/resilience-stages`, `/shu/attack-profitability`, `/shu/service-providers` in `sitemap.xml` — every one of these now 404s (`app/[daoId]/layout.tsx` calls `toDaoIdEnum(daoId)` → `null` → `notFound()` since `"SHU"` is no longer in `ALL_DAOS`).
- `llms.txt`'s "Key DAO pages" list keeps `{baseUrl}/shu` for the same reason.

Net effect: search engines and LLM crawlers keep being pointed at a dead page from the site's own sitemap/llms.txt, which is exactly the kind of stale-link cleanup this PR set out to do. Fix: derive `daoIds` in both files from `ALL_DAOS` (lowercased) instead of `Object.keys(daoConfigByDaoId)`.

Worth noting: this is a **pre-existing gap**, not new — `OP` has the identical problem today (also still a `rawDaoConfigByDaoId` key, also absent from `ALL_DAOS`), so `/op` is presumably already dead-linked in the current sitemap/llms.txt. Not asking this PR to fix the `OP` case too, just flagging that fixing it for `SHU` here (by switching to `ALL_DAOS`) would fix both DAOs at once. [Code-only]

**Not a bug (checked, no change needed):** `shared/utils/whitelabel.ts`'s `NORMALIZED_HOSTNAME_TO_DAO_ID` map has the same root cause (built from `daoConfigByDaoId`, still resolves `shutter.gov.blockful.io` → `"SHU"`), but it's harmless — a visit to that hostname gets rewritten internally to `/whitelabel/shu/...`, and `app/whitelabel/[daoId]/layout.tsx` re-validates via `toDaoIdEnum` and correctly `notFound()`s. End state is a 404 either way, just via one extra internal hop.

---

### `DaoOverviewHeader.tsx` — price-formatting drop (confirmed harmless)

`lastPrice.toFixed(daoId === DaoIdEnum.SHU ? 4 : 2)` → `lastPrice.toFixed(2)` (`apps/dashboard/features/dao-overview/components/DaoOverviewHeader.tsx:64`). Unlike the other touched files, this one drops the SHU special case outright rather than keeping it behind a commented/string-compared check. No live-user impact today since `/shu` (and the panel entry) are unreachable, so this component never renders for SHU anymore — flagging only so it's not missed if Shutter is ever re-enabled (the 4-decimal formatting would need to be manually re-added, unlike the other call sites which restore verbatim on uncomment). [Code-only]


---

PR #2142, opened 2026-08-25
fix: PR #2139 review feedback (Tornado voter filter, relayer receipt errors)

brunod-e:
Addresses the two open Codex review comments on #2139.

**`fix(dashboard)`: exclude the voter from Tornado delegated-vote addresses** (P1)
The delegators query intentionally returns self-delegations (the `TORNGovernor:Undelegated` handler records an account as its own delegate), but the governor forbids self-delegation, so a `from` list containing the connected wallet made `castDelegatedVote` revert and blocked the vote. The `from` list is now built by `buildTornDelegatedVoteFrom`, which drops the voter's own address case-insensitively (and keeps the existing dedup); a voter left with no delegators routes to `castVote`. Covered by unit tests.

**`fix(relayer)`: hold the enactment lock when the post-broadcast receipt wait throws** (P2)
When `sendTransaction` succeeded but the initial `waitForReceipt` threw a transient RPC error, `runEnactment` rejected and the rejection handler released the in-flight lock while the transaction could still be pending, so a retry could broadcast a duplicate. The failed poll is now treated as an unsettled broadcast: the caller still gets the tx hash, and the lock is handed to the existing bounded settlement hold (`releaseWhenSettled`). Covered by a new unit test.

### DIFF
--- a/.changeset/relayer-hold-lock-on-receipt-error.md
+++ b/.changeset/relayer-hold-lock-on-receipt-error.md
@@ -1,0 +1,5 @@
+---
+"@anticapture/relayer": patch
+---
+
+Keep the enactment lock held when the post-broadcast receipt wait fails with an RPC error.
--- a/.changeset/torn-exclude-voter-from-delegated-vote.md
+++ b/.changeset/torn-exclude-voter-from-delegated-vote.md
@@ -1,0 +1,5 @@
+---
+"@anticapture/dashboard": patch
+---
+
+Exclude the voter's own address from the Tornado delegated-vote list so castDelegatedVote does not revert.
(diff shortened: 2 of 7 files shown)


isadorable-png:
## 🎨 UI Review

> Automated review · Figma: not found · Spec: not found · Preview: dashboard preview not built for this PR (Vercel marked the `anticapture` project deployment "Ignored"); the only Ready preview (storybook) was unreachable from this environment.
> ℹ️ Reviewed without a Figma reference — no ClickUp task found either (searched PR body/comments, the two preconfigured ClickUp lists, and a workspace-wide ClickUp search for "Tornado"/"delegated vote"). Review grounded directly in the diff.

---

### VotingModal.tsx

No visible UI change. The diff only touches the body of the `tornDelegatedVoteAddresses` `useMemo` — it replaces ~15 lines of inline dedup/filter logic with a call to the new `buildTornDelegatedVoteFrom(address, ...)` utility. No JSX, markup, props, or render-affecting imports are touched: same modal, same layout, same states as before. Nothing to flag visually here.

`apps/relayer/src/services/proposals/proposal-enactment.ts` and the new test files (`tornDelegatedVote.test.ts`, `proposal-enactment.test.ts`) are backend/logic-only and out of scope for a UI review.

*(Note: the Tornado voting UX issue raised on #2139 — disabled Submit button with no loading/error explanation — is unrelated to this diff and isn't re-raised here.)*

---

PR #2145, opened 2026-08-28
hotfix: apply vote-by-voter filtering and stop feed events gateway timeouts

brunod-e:
## Context

Reported by the Uniswap Foundation via the MCP: vote-by-voter filtering was ignored on `votes`, and `feed/events` returned 5xx. Investigation found three distinct bugs, all reproduced against production.

## Fixes

1. **`GET /{dao}/votes` ignored `voterAddressIn`** — the handler accepted the param in its schema but never passed it to the query, silently returning unfiltered votes. Reproduced live: filtering by one voter returned other voters. The filter is now applied (`inArray` on `voter_account_id`), matching the by-proposal endpoint that already worked.

2. **`feed/events` 504 on address-filtered queries** — `buildAddressFilter` ran a correlated `EXISTS` probe against the source tables for every `feed_event` row, which blew past the 15s upstream timeout on large DAOs (reproduced on UNI: `?address=...` → HTTP 504 in 15.2s). The filter now uses uncorrelated `IN` subqueries (one scan per source table, hashed by Postgres) and only builds the branches for the requested event types, so a `type=VOTE` request never touches the transfers table.

3. **MCP dropped every array query param** — the MCP's upstream axios client used the default serializer, emitting `type[]=VOTE` / `voterAddressIn[]=0x...`, which the API does not parse. Verified against production: `type[]=VOTE` returned TRANSFER events. The client now sends repeated params (`type=VOTE&type=PROPOSAL`), confirmed parsed correctly by the production API.

Also: address-list query params (`voterAddressIn`, `addresses`, `delegateAddressIn`, ...) now accept the comma-delimited form their OpenAPI descriptions already promised (previously a 400).

## Verification

- Full API suite: 1078/1078 passing, including new tests for the `/votes` filter (single, repeated, and comma forms), comma splitting in the shared mapper, and address+type feed filtering.
- Typecheck and lint clean on `@anticapture/api` and `@anticapture/client`.
- Axios serializer behavior verified directly (bracketed vs repeated output).
- After deploy, re-check the prod repro: `GET /uni/feed/events?address=0x8962...&limit=100` (was 504) and `GET /uni/votes?voterAddressIn=0x8962...` (was unfiltered).

No OpenAPI schema changes, so no client regen needed. Changesets included (3× api patch, 1× client patch).

### DIFF
--- a/.changeset/address-query-comma-lists.md
+++ b/.changeset/address-query-comma-lists.md
@@ -1,0 +1,5 @@
+---
+"@anticapture/api": patch
+---
+
+Accept comma-delimited address lists in query params as documented.
--- a/.changeset/feed-address-filter-timeout.md
+++ b/.changeset/feed-address-filter-timeout.md
@@ -1,0 +1,5 @@
+---
+"@anticapture/api": patch
+---
+
+Speed up the feed events address filter to stop gateway timeouts on large DAOs.
--- a/.changeset/mcp-array-params.md
+++ b/.changeset/mcp-array-params.md
@@ -1,0 +1,5 @@
+---
+"@anticapture/client": patch
+---
+
+Send array query params as repeated params from the MCP server so filters are applied.
--- a/.changeset/votes-voter-filter.md
+++ b/.changeset/votes-voter-filter.md
@@ -1,0 +1,5 @@
+---
+"@anticapture/api": patch
+---
+
+Apply the voterAddressIn filter on the all-votes endpoint.
--- a/apps/api/src/controllers/votes/onchainVotes.integration.test.ts
+++ b/apps/api/src/controllers/votes/onchainVotes.integration.test.ts
@@ -305,6 +305,51 @@ describe("Onchain Votes Controller", () => {
       const body = await res.json();
       expect(body).toEqual({ items: [FULL_VOTE_OBJECT], totalCount: 1 });
     });
+
+    it("should accept voterAddressIn filter", async () => {
+      await db.insert(proposalsOnchain).values(createProposal({ id: "1" }));
+      await db.insert(votesOnchain).values([
+        createVote({ proposalId: "1", voterAccountId: VOTER_ADDRESS }),
+        createVote({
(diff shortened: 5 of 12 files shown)


brunod-e:
## Verification against real UNI data

Ran the branch locally against the dev database (real UNI dataset: 9.6M feed_event rows, 9.2M transfers) and exercised the exact repros from the UF report:

| Query | Before (prod / branch pre-fix) | After |
|---|---|---|
| `feed/events?address=0x8962...` (default relevance) | 504 in 15s / statement timeout 30s | **200 in 1.4s** |
| `feed/events?address=0x8962...&relevance=ALL` | 504 in 15s | **200 in 1.5s** |
| UF repro: `address=0x3FB1...&relevance=ALL&fromDate=1784000000` | timeout via MCP | **200 in 0.5s** — 3 events (1 DELEGATION, 2 VOTE) |
| same + `type=DELEGATION` | filter silently dropped via MCP | **only the DELEGATION event** |
| Binance address + `relevance=ALL` (thousands of transfers) | — | 200 in 2.1s |
| `votes?voterAddressIn=0x8962...` | returned other voters | **24 votes, all by that voter** |

EXPLAIN ANALYZE on the dev DB drove the last commit: `LOWER()` comparisons couldn't use the account indexes (transfers subquery became a per-row rescanned seq scan, cost 383k → 4.4k as an indexed BitmapOr), and the LIMIT-driven timestamp index walk scanned all 9.6M entries when an address had fewer matches than the page size. Addresses are stored viem-checksummed (sampled 50k transfer rows: 49,990 checksummed, 10 digit-only), so equality against both written forms is safe and indexed.

Result contents validated: only events involving the filtered address, descending timestamp order preserved, existing no-address feed queries unchanged (1.1s baseline).

Note: the MCP server itself only starts sending arrays correctly once the `@anticapture/client` fix is released and the MCP deployment picks it up — until then array filters through the MCP stay dropped even with the API fixed.

PR #2147, opened 2026-09-01
fix: aave stakeholders timeouts and missing address column

brunod-e:
## Problem

[DEV-1167](https://app.clickup.com/t/86ak8ht2h): the Aave stakeholders page shows no addresses in production. Two independent causes:

1. **Backend**: `/aave/voting-powers` and `/aave/balances` take 6-15s cold and regularly exceed Gateful's upstream timeout, surfacing as 504 and leaving the table empty (reproduced via the dashboard proxy; ENS/UNI answer in ~0.3s).
2. **Frontend**: even when the data loads, at desktop widths (≥1024px) the delegates table renders avatars with no address text — reproduced on production at 1600px and confirmed in the DOM: the name span is present with the right text but 0px wide.

## Cause

**API** — the Aave-specific repositories aggregate the whole dataset on every request before paginating 20 rows:

- `voting-power/aave.ts` built a UNION of all ~788k account ids, a `SUM(balance)` over the entire `account_balance` table and a `SUM(delta)` over the windowed `voting_power_history`, joined them all, then ordered and took a page. The `COUNT` query repeated the same heavy joins (including one it never used) and ran sequentially after the page query.
- `account-balance/aave.ts` aggregated every transfer in the 90d window for every account even when ordering by balance, and its `COUNT` dragged the same transfer aggregation in for nothing.

**Dashboard** — the Aave-only `DelegationTable` copy dropped the `min-w-0 flex-1` wrapper around `EnsAvatar` and pinned the address column to `w-40` (160px). The hover controls (copy + Details) render with `opacity-0` but still occupy ~140px, so the name span (inside `min-w-0` + `overflow-hidden`) collapses to 0. Below `lg` the controls don't render (`isMobile`), which is why the bug only shows on desktop.

## Fix

- **API**: select the page of accounts first, joining a full-table aggregation only when the requested ordering depends on it, then compute the remaining per-account values for just the returned page through the indexed `account_id` columns. Count queries drop the aggregation joins they never needed and run in parallel with the page query. Also fixes a latent bug (`signedVariation` applied `desc()` twice, producing invalid SQL) and makes pagination stable with an `account_id` tiebreaker.
- **Dashboard**: mirror the shared `Delegates` address cell in the Aave copy — flexible address column, `w-40 lg:w-full` wrapper, and the name claiming the remaining space via `min-w-0 flex-1`.

## Verification

- Reproduced both symptoms beforehand: the 504 on `/aave/balances` plus 6-11s cold latencies, and the 0px-wide name span on production at 1600px.
- Full API suite (1078 tests), typecheck and lint pass; the Aave repository suites run the real SQL against PGlite.
- A throwaway old-vs-new benchmark on a seeded PGlite (75k balances, 15k powers, 120k history rows, 100k transfers) returned identical results for every ordering with the default views dropping 496ms → 199ms (delegates) and 389ms → 107ms (holders); gains should be larger on prod-size data since the old cost grows with table size and the new one is page-scoped.
- The dashboard fix reuses the exact cell structure of the shared `Delegates`/`TokenHolders` tables, verified working on production at 1600px (`nick.eth` renders with a flexible 330px cell on ENS). Dashboard typecheck and lint pass.

### DIFF
--- a/.changeset/aave-delegates-address-column.md
+++ b/.changeset/aave-delegates-address-column.md
@@ -1,0 +1,5 @@
+---
+"@anticapture/dashboard": patch
+---
+
+Show the address column again on the Aave delegates table at desktop widths.
--- a/.changeset/aave-stakeholders-timeout.md
+++ b/.changeset/aave-stakeholders-timeout.md
@@ -1,0 +1,5 @@
+---
+"@anticapture/api": patch
+---
+
+Fix the Aave stakeholders page timing out by scoping the voting-power and balance listing aggregations to the requested page.
--- a/apps/api/src/repositories/account-balance/aave.ts
+++ b/apps/api/src/repositories/account-balance/aave.ts
@@ -113,49 +113,121 @@ export class AAVEAccountBalanceRepository {
       amountfilter,
     );
 
-    const variations = this.queryFragments.variationCTE(
-      variationFromTimestamp,
-      variationToTimestamp,
-      filter,
-    );
-
-    const [totalCount] = await this.db
-      .select({
-        count: sql<number>`COUNT(DISTINCT ${variations.accountId})`.as("count"),
-      })
-      .from(variations);
-
-    const aggregated = this.db
-      .select({
-        accountId: variations.accountId,
-        delegate: sql<string>`MAX(${variations.delegate})`.as("delegate"),
-        currentBalance: sql<bigint>`SUM(${variations.currentBalance})`.as(
-          "current_balance",
-        ),
-        absoluteChange:
-          sql<string>`MAX(${variations.fromChange}::numeric) + MAX(${variations.toChange}::numeric)`.as(
-            "absolute_change",
-          ),
-      })
-      .from(variations)
-      .groupBy(variations.accountId)
-      .as("aggregated");
-
     const orderDirectionFn = orderDirection === "desc" ? desc : asc;
 
-    const orderByCriteria =
-      orderBy === "balance"
-        ? aggregated.currentBalance
-        : orderBy === "signedVariation"
-          ? sql`${aggregated.absoluteChange}::numeric`
-          : sql`ABS(${aggregated.absoluteChange}::numeric)`;
+    // Counting distinct accounts does not need the transfer aggregation the
+    // old variation CTE dragged in: the per-account transfer sums never add
+    // or remove accounts.
+    const countQuery = this.db
+      .select({
+        count: sql<number>`COUNT(DISTINCT ${accountBalance.accountId})`.as(
+          "count",
+        ),
+      })
+      .from(accountBalance)
+      .where(filter);
 
-    const result = await this.db
-      .select()
-      .from(aggregated)
-      .orderBy(orderDirectionFn(orderByCriteria))
-      .offset(skip)
-      .limit(limit);
+    // Aave holds hundreds of thousands of balances, so aggregating every
+    // transfer in the window for every account on every request times out
+    // the gateway. Group balances per account first; join the windowed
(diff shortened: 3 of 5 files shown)


isadorable-png:
## 🎨 UI Review

> Automated review · [Figma: 🛰️ Product Design — Holders & Delegates (Shipped)](https://www.figma.com/design/mUgy2KpQ3gJ07yZaUaXu8l/%F0%9F%9B%B0%EF%B8%8F-Product-Design?node-id=6-95305) · [Spec: DEV-1167](https://app.clickup.com/t/86ak8ht2h) · Preview: blocked by network egress on this run (not auth — the Vercel URL just isn't reachable from here), so this is grounded in Figma + a direct code diff against the already-shipped, production-verified shared `Delegates`/`TokenHolders` tables rather than a live screenshot.
>
> Note on the Figma link: the node stored on the DEV-1167 ClickUp task (`node-id=1385-30140`) no longer exists in the file. I located the equivalent current design — the `Table Row / Delegates` component set and `Desktop / DAO details - Holders & Delegates` frame under the file's "Shipped" page — and used that instead.

---

### Aave Stakeholders — Desktop (`DelegationTable.tsx`)

**Address column regression fix**
- Validated against Figma — no change needed. The new wrapper (`group flex w-40 items-center lg:w-full` → `min-w-0 flex-1` around `EnsAvatar`) is a byte-for-byte match of the shipped `Delegates.tsx` address cell (`apps/dashboard/features/holders-and-delegates/delegate/Delegates.tsx`), which mirrors the `Table Row / Delegates` component set in Figma (avatar + dashed underline name, hover-only copy/details actions). No further change needed here. [Figma-confirmed]

**New: tablet-range regression (769–1023px)** — same root cause the PR fixes, narrower window
- `apps/dashboard/app/aave/stakeholders/DelegationTable.tsx:222` — the column is pinned to `w-40` (160px) until Tailwind's `lg` breakpoint (1024px, no override in `apps/dashboard/app/globals.css`), but the hover-controls sibling (`CopyAndPasteButton` + `Details` button, ~140px, rendered with `opacity-0` so it still occupies layout width) is gated on `!isMobile`, and `isMobile` flips to `false` at `TABLET_SCREEN_SIZE = 768` (`apps/dashboard/shared/constants/screen-size.ts`, checked via `matchMedia(max-width: 768px)` in `apps/dashboard/shared/hooks/useScreenSize.tsx`). Between ~769px and 1023px the controls render inside a column still capped at 160px — reproducing the exact 0-width name collapse this PR fixes, just at a narrower width than the 1600px case verified in the PR description.
- This isn't introduced by this PR — the shared `Delegates.tsx:265` (`apps/dashboard/features/holders-and-delegates/delegate/Delegates.tsx`) this diff mirrors has the identical breakpoint mismatch and would show the same collapse in that range today.
- Fix: use the same breakpoint on both sides — swap `lg:w-full` for `md:w-full` (Tailwind's default `md` = 768px, matching `TABLET_SCREEN_SIZE`), or drive the column width off `isTablet`/`isDesktop` from `useScreenSize` instead of a separate CSS breakpoint.
- **Question for author**: worth fixing here too (2-line change, same file), or is a follow-up ticket preferred since the shared component has the same latent bug and is out of this diff's scope? [Code-only]

---

### Mobile

No mobile-specific regression from this diff — below 768px, `isMobile` is `true` so the hover-controls sibling never renders and the address column keeps its full width for the name. The only gap is the 769–1023px tablet band called out above. [Code-only]

---

### Backend (`apps/api`)

Out of scope for a UI review — no rendering/styling change, skipped per the review's scope.


---

pikonha (approved):

PR #2164, opened 2026-09-15
Dev

LeonardoVieira1630:


isadorable-png:
## 🎨 UI Review

> Automated review · Figma: [Anticapture Product Design](https://www.figma.com/design/mUgy2KpQ3gJ07yZaUaXu8l) — sections "🔍 Calldata Visualization (Decoder + AI Summary)" (node `3412:136491`) and "📬 Proposal Improvements (Creation Flow)" (node `3569:136267`), found via the linked ClickUp specs · Spec: [Calldata Visualization](https://app.clickup.com/t/86aje4uvc) · [Proposal Improvements](https://app.clickup.com/t/86aje4vfh) · Preview: not reachable (see caveat)
>
> ⚠️ **Mixed-coverage caveat.** This PR is a `dev`→`main` bulk sync (162 files, 105 commits), not a single feature. Two of the touched surfaces — the new Calldata Decoder and the field-message pattern in the create-proposal modals — matched a Figma spec found via ClickUp, so those sections below are Figma-validated. The rest (the gasless queue/execute flow in `GovernanceActionModal`, the sidebar/mobile nav entry) had no matching spec in the two ClickUp Features lists or a workspace search, so those are `[Code-only]`/`[UX-heuristic]`, grounded in diff reading and UX judgment.
> ⚠️ **Preview unreachable.** Vercel skipped the dashboard's own PR preview for this custom `dev` environment (shows "Ignored" in the bot comment); only `anticapture-storybook.vercel.app` deployed, and that was blocked by this environment's egress proxy. No live before/after comparison was possible — regression check below is diff-level only.

---

### Calldata Decoder (`/tools/decoder`, embedded Actions tab)

**Validated against Figma — no change needed**
- Field-message color mapping (`DecoderInputPanel.tsx`'s `HelperText`, `AbiInput.tsx`'s error span): helper → `text-secondary`, warning → `text-warning`, error → `text-error`. Matches Figma "01 — Field message component" (node `3569:136268`) exactly, including "error also sets the border, warning never touches the border" (`AbiInput`'s `Textarea` gets `error={Boolean(error)}` only on the error path). [Figma-confirmed]
- Empty / decoding / ABI-unknown / error state machine (`DecoderTool.tsx` + `useDelayedFlag`): raw hex never hidden, 300ms skeleton threshold before showing a loading state, AI/summary layer absent rather than shown broken when no trusted decode exists. Matches Figma node `3412:136498` state-by-state. [Figma-confirmed]
- `DecodedActionCard` / `ChipCluster` / `AddressChip` / `CollapsedActionRow` structure (header chip cluster with ABI-source badge, target identity chip, human-first param rows with type badge + humanized value + raw annotation, `Decoded | Raw` segmented footer) matches the "IN PRODUCTION" reference frame (node `3460:21732`) closely, down to the `//action NN` mono label style (`MONO_LABEL` constant is explicitly documented as "Figma `Alternative/xSmall`"). [Figma-confirmed]
- Standalone tool's empty state ("Nothing to decode yet" / "Paste the input data of any transaction…") uses different copy than the Figma empty state ("NO CALLDATA YET" / "Paste hex above, or open any proposal action."). This is a justified adaptation, not a bug: the Figma copy assumes the embedded-in-a-proposal context ("open any proposal action"), which doesn't exist on the standalone `/tools/decoder` route — the shipped copy is more accurate for its actual context. No change needed.

**Nice-to-have — component reuse**
- `DecoderInputPanel.tsx` defines a local `HelperText` component that reimplements the exact "shared field-message component" the linked spec calls for as a to-be-built DS primitive ("One shared field-message component with reserved space, used everywhere an error or warning appears in the flow"). The same helper/warning/error pattern also appears inline in `AddTransferModal.tsx` and `AddCustomActionModal.tsx`. Worth promoting `HelperText` to `shared/components/design-system/form/field-message/FieldMessage` so all three call sites (and future ones) share one implementation instead of three copies of the same three-color mapping. [Code-only]

**Nice-to-have — touch target**
- `ExpandToggle`'s chevron button is `size-5` (20×20px), below the 44×44px recommended minimum touch target. In `CollapsedActionRow` this is mitigated by the whole row being one large click target (`role="button"` with `p-3`), and in `SubcallList` the adjacent label text shares the same expand handler and fills the remaining row width — so it's tappable in practice, but a reader aiming precisely at the chevron on a small screen has a small target. [UX-heuristic]

### Governance action modal — gasless queue/execute (`GovernanceActionModal.tsx`)

No matching Figma spec was found for this flow (not covered by either ClickUp task this review located).

**Must-fix — color semantics**
- The "idle" entry state — `Checking whether this action can be sponsored...` — renders `<SpinIcon className="text-warning size-4 animate-spin" />`. `text-warning` signals caution/an issue; this is a neutral loading check with no problem to flag yet. Suggest `text-secondary` (matching the copy's own color) so a plain "still loading" moment doesn't read as a warning state. (`features/governance/components/modals/GovernanceActionModal.tsx`) [UX-heuristic]

**Validated — good pattern reuse**
- `GaslessBadge` (`bg-success/80 text-inverted`, extracted into `ProposalHeader.tsx`) consolidates a `"Free"` badge className that was previously duplicated verbatim in `DelegateButton.tsx` and the pre-PR `ProposalHeader.tsx` — a genuine de-duplication, not a new one-off. No change needed.

### Navigation (`HeaderSidebar.tsx` / `HeaderMobile.tsx`)

- New "Decoder" entry (icon: `Braces`, path `tools/decoder`, `isGlobal: true`) is added identically to both desktop sidebar and mobile header, with matching icon and label — good parity. `ButtonHeaderSidebar.tsx`'s active-state matcher was also updated to full-pathname matching for global pages, so a nested global route like `tools/decoder` highlights correctly instead of falling through to the DAO-page last-segment matcher. No issues found.

---

### Mobile

- `TabsSection.tsx`'s sticky-tabs offset was corrected from `top-29.5` to `top-13.75` with a code comment explaining the fixed mobile header is 55px, not the old proposal-header height — this fixes exactly the "sticky element pinning under the wrong offset" class of bug called out in this review's own checklist. [Code-only, confirmed via the diff's own before/after and comment]
- `DecoderInputPanel`'s three ABI-source tabs scroll horizontally on narrow viewports instead of wrapping or overflowing the card (`overflow-x-auto` with a code comment acknowledging the phone-width constraint) — correct mobile pattern. No issues found.

---

### Findings summary

3 substantive findings — 1 must-fix (idle spinner color), 2 nice-to-have (shared `FieldMessage` component, `ExpandToggle` touch target) — plus 4 items validated against Figma with no change needed. The decoder feature (the bulk of this PR's new UI surface) is unusually well-built: consistent design-token usage throughout (`text-primary`/`text-secondary`/`text-warning`/`text-error`/`border-border-*`), no hardcoded hex colors, and structure that matches the linked Figma spec closely enough that most "looks different" candidates turned out to be intentional (see "Validated" sections above).

---

brunod-e (approved):

